PR #2, opened 2021-01-22
2D semi-simplicial sets

epatters:
This provides the basic data structure and interface. Next up should be orientation, chains and cochains, and the boundary and coboundary (exterior derivative) operators.

### DIFF
--- a/src/SimplicialSets.jl
+++ b/src/SimplicialSets.jl
@@ -42,10 +42,10 @@ const SimplexCategory1D = TheoryReflexiveGraph
 """
 const AbstractSemiSimplicialSet1D = AbstractGraph
 
-""" A 1D semi-simplicial set.
+""" A one-dimensional semi-simplicial set.
 
-One-dimensional semi-simplicial sets are the same as graphs, and this type is
-just an alias for `Graph`. The boundary operator [`∂₁`](@ref) translates the
+Semi-simplicial sets in 1D are the same as graphs, and this type is just an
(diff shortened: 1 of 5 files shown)


src/SimplicialSets.jl
```diff
@@ -0,0 +1,160 @@
...
+end
+
+# 2D simplicial sets
+####################
+
+@present SemiSimplexCategory2D <: SemiSimplexCategory1D begin
+  Tri::Ob
+  src2_first::Hom(Tri,E) # ∂₂(2)
+  src2_last::Hom(Tri,E)  # ∂₂(0)
+  tgt2::Hom(Tri,E)       # ∂₂(1)
```
jpfairbanks, line 80: these names aren't very intuitive. if they are standard, we could provide an ASCII art diagram in this file.

epatters: I will add some docs explaining the names.

I had originally called these fields `edge2`, `edge0`, and `edge1`, which is arguably more standard but is even less intuitive!

jpfairbanks: Edge 0-2 at least is completely arbitrary so you just have to memorize what it means. The src first src last is like trying to decode a little puzzle where you think it should be intuitive. As ping as we have docs it doesn't really matter.

epatters:
Going to merge this now. See additional docs. Open to changing the field names if we can come up with better ones.

PR #5, opened 2021-01-26
Physics System Examples

bosonbaas:
This PR adds the example of a spring system. This spring system will take advantage of the boundary/coboundary operators to allow for spring systems to be evaluated on an arbitrary mesh.


jpfairbanks:
This code went into Decapodes.jl

PR #15, opened 2021-02-24
Importing and visualizing meshes

bosonbaas:
This PR adds tooling in `src/Visualization.jl` to allow for importing .obj files into DeltaSets and to support visualization of DeltaSets and their Duals.


bosonbaas:
I think this PR is now ready for review! I'd appreciate thoughts on
- visualization file name
- whether import tooling should be put somewhere else
- how to deal with dependencies (AbstractPlotting still depends on a lot of packages)

jpfairbanks:
I think with plotting, we need to use Requires.jl until conditional dependencies are more directly handled. Drawing the meshes is going to be so core to debugging, that I find it hard to believe that anyone will be able to use CombinatorialSpaces without visualizing stuff, which leads me to believe that it is ok to depend on the plotting ecosystem even if it is big.

bosonbaas:
Do we have any examples of using Requires.jl? I haven't worked with that package before

bosonbaas:
Would it make sense to have a MeshUtils.jl file which contains the tooling for:
- Importing mesh files
- Converting meshes to DeltaSets
- Converting DeltaSets to meshes
?
This would just be migrating the current tooling out of Visualization.jl.

src/MeshUtils.jl
```diff
@@ -0,0 +1,70 @@
...
+import ..SimplicialSets: EmbeddedDeltaSet2D
+
+export make_mesh, EmbeddedDeltaSet2D
+
+# Helper Functions (should these be exposed?)
+##################
+
+""" Construct a GeometryBasics.Mesh object from an embedded delta set
+"""
+function make_mesh(ds::EmbeddedDeltaSet2D)
```
epatters, line 24: For symmetry with the constructor `EmbeddedDeltaSet2D(::Mesh)` below, we could replace `make_mesh` with a new constructor `Mesh(::EmbeddedDeltaSet2D)`. My understanding is that this is *not* type piracy, and now you don't need to worry whether to export another function.

src/MeshUtils.jl
```diff
@@ -0,0 +1,70 @@
...
+  points = Point{3, Float64}[point(ds)...]
+  tris = TriangleFace{Int}[zip(triangle_vertices(ds)...)...]
+  GeometryBasics.Mesh(points, tris)
+end
+
+""" Construct a GeometryBasics.Mesh object from a dual embedded delta set
+"""
+function make_mesh(ds::EmbeddedDeltaDualComplex2D)
```
epatters, line 32: Since the dual complex also includes the data of the primal complex, does it make sense that this function always uses the dual triangles?

src/MeshUtils.jl
```diff
@@ -0,0 +1,70 @@
...
+  coords = coordinates(m)
+  tris = faces(m)
+  s = EmbeddedDeltaSet2D{Bool, eltype(coords)}()
+  add_vertices!(s, length(coords), point=coords)
+  for tri in tris
+    tri = convert.(Int64, tri)
+    glue_sorted_triangle!(s, tri[1], tri[2], tri[3])
+  end
+  # Assign orientation to 1, then spread to neighbors
+  orient_component!(s, 1, true)
```
epatters, line 56: This can fail to orient the complex in two ways:

- because the mesh is not orientable at all
- because the mesh has multiple connected components

Do we need to worry about that?

src/Visualization.jl
```diff
@@ -0,0 +1,55 @@
...
+This module provides wrapper functions for plotting embedded delta sets
+using the Makie functionality. This includes the wireframe, mesh, and scatter
+plotting functions.
+
+It is important to note that the end-user will need to import one of the Makie
+backends (CairoMakie, GLMakie, or WGLMakie) in order for plotting to work
+correctly.
+"""
+
+module Visualization
```
epatters, line 12: The name `Visualization` isn't bad, but since you asked about what to call this module, another, more specific name would be `MeshGraphics`.

epatters (commented):
Thanks Andrew, this is great! I left a few comments.

src/MeshUtils.jl
jpfairbanks: Yes, this is not piracy because we "own" `EmbeddedDeltaSet2D`

src/MeshUtils.jl
jpfairbanks: For multiple components, we should find a deterministic way to pick an ordering for each connected component. Raising a clear exception in the case of unorientable complex seems fine. We generally don't try/catch things and just let stack traces propagate. Making sure the leaf exception is intelligible should be sufficient.

epatters: I agree. OK, so what I am going to do is add a higher-level that functions that finds the connected components and then orients them, and then we can update this code. So don't worry about this for now, Andrew.

src/MeshUtils.jl
bosonbaas: Yeah, that's a good point. I made this function primarily for visualizing the dual triangles, which is why it only uses those.
Would it make sense to add a flag for meshing either the primal or dual triangles (defaulting to dual) when passed a DualComplex?

bosonbaas:
Thanks for the comments! I've gotten most of them fixed, and responded to the ones I haven't fixed yet.

src/MeshUtils.jl
epatters: Yes, I think adding that keyword argument would make sense.

src/MeshUtils.jl
epatters: This improvement is now implemented in PR #18.

epatters:
I made the asset paths more portable to make local testing easier.

When I build the Documeter.jl docs locally, I don't see any graphics. Is this working for you?

bosonbaas:
Thanks for catching that, I had forgotten to change filenames in the docs. You should be able to see some interactive 3D cats in the docs now!

epatters:
I can confirm that this now works in the browser. So dang cool.

Thanks @bosonbaas and @olynch! Merging now.

PR #16, opened 2021-02-25
Consistently orient simplices in a connected component

epatters:
@bosonbaas, this should be useful for you.


bosonbaas:
Yeah, this looks great! Excited about the level of generality this'll give us!

jpfairbanks:
I'm excited for the obvious benefits we get from ACSets, we can store these things with their orientations, which should save time in the mesh generation phase of a simulator! One of our avenues for speedup is that we can avoid certain computations entirely because we have better data structures than Matrix.

PR #24, opened 2021-04-05
Mesh support for STL files

bosonbaas:
With the current mesh tooling, STL formatted mesh files are imported with each triangle disconnected from its neighbors. 
This PR merges these meshes pointwise (points at the same place are merged into a single point) in the conversion to delta sets.

### DIFF
--- a/src/MeshInterop.jl
+++ b/src/MeshInterop.jl
@@ -48,13 +48,20 @@ end
 This operator should work for any triangular mesh object. Note that it will
 not preserve any normal, texture, or other attributes from the Mesh object.
 """
-function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh)
+function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh; force_unique=false)
   coords = coordinates(m)
+  ind_map = collect(1:length(coords))
+  if(force_unique) 
+    indices = unique(i->coords[i], 1:length(coords))
+    val2ind = Dict(coords[indices[i]]=>i for i in 1:length(indices))
+    ind_map = map(c->val2ind[c], coords)
+    coords = coords[indices]
+  end
   tris = faces(m)
   s = EmbeddedDeltaSet2D{Bool, eltype(coords)}()
   add_vertices!(s, length(coords), point=coords)
   for tri in tris
-    tri = convert.(Int64, tri)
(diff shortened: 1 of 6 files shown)


bosonbaas:
This PR has also changed the graphing backend from AbstractPlotting to Makie. Apparently this transfer was made by the Makie community in their last version update.

src/MeshInterop.jl
```diff
@@ -69,6 +76,10 @@ that it will not preserve any normal, texture, or other data beyond points and
 triangles from the mesh file.
 """
 function EmbeddedDeltaSet2D(fn::String)
-	EmbeddedDeltaSet2D(FileIO.load(fn))
+  if(fn[(end-2):end] == "stl")
```
epatters, line 79: Should use the `splitext` function to get the file extension.

src/MeshInterop.jl
```diff
@@ -48,13 +48,20 @@ end
 This operator should work for any triangular mesh object. Note that it will
 not preserve any normal, texture, or other attributes from the Mesh object.
 """
-function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh)
-  coords = coordinates(m)
+function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh; force_unique=false)
+  coords = metafree.(coordinates(m))
+  ind_map = collect(1:length(coords))
```
epatters, line 53: I could be wrong, but I doubt you need a `collect` here.

epatters (commented):
Thanks Andrew! Minor comments below.

src/MeshInterop.jl
```diff
@@ -48,13 +48,20 @@ end
 This operator should work for any triangular mesh object. Note that it will
 not preserve any normal, texture, or other attributes from the Mesh object.
 """
-function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh)
-  coords = coordinates(m)
+function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh; force_unique=false)
```
epatters, line 51: Probably worth documenting what the keyword argument `force_unique` does and why you need it for STL files.

bosonbaas:
Thanks for the feedback! Those should be taken care of in this latest commit.

epatters:
Thanks! Merging.

PR #28, opened 2021-05-20
Discrete primal-primal sharp operator

epatters:
Following (Hirani, 2003) and (Desbrun et al, 2005) but see the caveats in the docstring.


epatters:
Thanks for your help debugging this, @bosonbaas. I think the new code should be equivalent to your working code, but can you please confirm?

bosonbaas:
I've run yours alongside mine, and they seem to be equivalent! 
The image with less mesh dependence that I showed before came from dividing the form by the length of the edge it corresponds with. I think this might relate to the 1/h term from the divergence of the interpolation function, but this is something that I think can be added later!

epatters:
OK, thanks. I'll go ahead and merge this but I'll look into that factor of 1/h.

PR #29, opened 2021-05-26
Sharp Operator Update

bosonbaas:
This PR slightly modifies the existing sharp operator so that it more closely aligns with Hirani's definition of the primal-primal sharp operator.

### DIFF
--- a/src/DualSimplicialSets.jl
+++ b/src/DualSimplicialSets.jl
@@ -448,9 +448,9 @@ function ♯(s::AbstractDeltaDualComplex2D, α::AbstractVector, ::PPSharp)
     tri_center, tri_edges = triangle_center(s,t), triangle_edges(s,t)
     tri_point = dual_point(s, tri_center)
     for (i, (v₀, e₀)) in enumerate(zip(triangle_vertices(s,t), tri_edges))
-      e_vec = s[s[e₀, :tgt],:point] - s[s[e₀, :src],:point]
+      e_vec = point(s, tgt(s, e₀)) - point(s, src(s, e₀))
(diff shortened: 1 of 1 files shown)


src/DualSimplicialSets.jl
```diff
@@ -448,13 +448,17 @@ function ♯(s::AbstractDeltaDualComplex2D, α::AbstractVector, ::PPSharp)
     tri_center, tri_edges = triangle_center(s,t), triangle_edges(s,t)
     tri_point = dual_point(s, tri_center)
     for (i, (v₀, e₀)) in enumerate(zip(triangle_vertices(s,t), tri_edges))
-      out_vec = tri_point - dual_point(s, edge_center(s,e₀))
+      e_vec = s[s[e₀, :tgt],:point] - s[s[e₀, :src],:point]
```
epatters, line 451: We should use the `src`, `tgt`, and `point` accessor functions here and below.

epatters (commented):
Thanks Andrew! Can you please update the docstring for the sharp operator later in the file, which is now out of date?

bosonbaas:
Thanks for the feedback! I've updated the docstrings, but I'm wondering if we still want to include the caveat (since we aren't using the Desbrun description). 

At this point I think it makes sense to remove it and just state that this is an implementation of Hirani's PP sharp (as referenced), but I wanted to get your thoughts on that before removing it in case you think it's an important clarification to make.

epatters:
Thanks Andrew. Yes, I will further clarify the docstring but I want to leave a note that this differs from Desbrun et al's formula to avoid someone getting confused about that later.

PR #31, opened 2021-06-11
DEC Operators for Barycentric Subdivision

bosonbaas:
These are some edits to the hodge star and flat operator in order to account for dual meshes which are generated with barycentric subdivision.

This is a WIP and will need significant cleanup, documentation, and tests. Currently this should reduce to the original operators for circumcentric subdivision.


epatters:
There are lots of possibilities for setting up a flat operator (Hirani suggests 8). The one Hirani most discusses is a dual-primal flat (sends dual vector fields to a primal 1-forms), but if it's also convenient to have a primal-primal flat, we could have that too. No need to restrict to only one kind of flat.

bosonbaas:
This PR is old and was never completed. It would probably be better to start over again on these barycentric subdivision operators, so an issue has been filed #52

PR #32, opened 2021-06-15
GAT for exterior calculus

epatters:
GATs for the exterior calculus specialized to dimensions 1 and 2.

### DIFF
--- a/src/ExteriorCalculus.jl
+++ b/src/ExteriorCalculus.jl
@@ -3,9 +3,10 @@ export Ob, Hom, dom, codom, compose, ⋅, id,
   otimes, ⊗, munit, braid, oplus, ⊕, mzero, swap,
   mcopy, Δ, delete, ◊, plus, +, zero, antipode,
   MetricFreeExtCalc1D, MetricFreeExtCalc2D, ExtCalc1D, ExtCalc2D, FreeExtCalc2D,
-  Space, Chain0, Chain1, Chain2, Form0, Form1, Form2, ∂₁, ∂₂, d₀, d₁,
-  DualForm0, DualForm1, DualForm2, dual_d₀, dual_d₁, δ₁, δ₂,
-  ⋆₀, ⋆₁, ⋆₂, ⋆₀⁻¹, ⋆₁⁻¹, ⋆₂⁻¹
+  Space, VectorField, Chain0, Chain1, Chain2, Form0, Form1, Form2,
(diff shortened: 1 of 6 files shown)


src/ExteriorCalculus.jl
```diff
@@ -0,0 +1,141 @@
...
+product that is bi-additive, i.e., additive in each variable separately.
+Sometimes these are called "tensor categories" but that term is ambiguous, as
+noted at the nLab. Not to be confused with an SMC that happens to be written
+additively (`SymmetricMonoidalCategoryAdditive` in Catlab).
+
+TODO: Migrate to `Catlab.Theories.MonoidalMultiple`.
+"""
+@theory AdditiveMonoidalCategory{Ob,Hom} <: AdditiveCategory{Ob,Hom} begin
+  # Tensor product.
+  otimes(A::Ob, B::Ob)::Ob
```
jpfairbanks, line 27: How do the otimes and oplus here relate to the tensor product and direct sum on the forms?

src/ExteriorCalculus.jl
```diff
@@ -0,0 +1,141 @@
...
+    (A::Ob, B::Ob, C::Ob, D::Ob, f::(A → B))
+end
+
+# Metric-free exterior calculus
+###############################
+
+""" Theory of exterior caclulus on 1D manifold-like spaces.
+"""
+@theory MetricFreeExtCalc1D{Ob,Hom,Space} <: AdditiveMonoidalCategory{Ob,Hom} begin
+  Space::TYPE
```
jpfairbanks, line 63: should we relate the coproduct on spaces to the direct sum of their cochains here?

src/ExteriorCalculus.jl
epatters: The general operations yield the corresponding operations on forms since forms live in particular objects of this category, e.g., `Form0(X)::Ob` and `Form1(X)::Ob`. Note that particular forms are modeled as generalized elements, so that a 1-form has type `Hom(munit(), Form1(X))`.

src/ExteriorCalculus.jl
epatters: Yep, we should add that. In practice, I assume spaces will usually be glued together via pushout so we'll need something more general.

epatters:
The main things missing are the Laplace-Beltrami and Laplace-de Rahm operators, but I also need to extend the DEC code to support the latter, so I'll handle all in that in a later PR.

PR #33, opened 2021-06-18
Laplace-de Rham operator

epatters:
Implements in the Laplace-de Rham operator for the DEC and adds both the Laplace-Beltrami and Laplace-de Rham operators to the exterior calculus GAT.


epatters:
@bosonbaas, warning that there two breaking changes here: I now use `Δ` for Laplace-de Rham, rather than Laplace-Beltrami, and I changed the sign convention for Laplace-Beltrami.

bosonbaas:
Thanks for the heads up, glad to have this functionality added! Currently my tooling in AlgebraicDynamics uses its own implementation of Laplace-de Rham, so it'll be nice to use this one (and this doesn't end up being a breaking issue for that tooling)

PR #35, opened 2021-07-22
Updated Hodge Star for 1-Forms

bosonbaas:
This PR provides an implementation of the hodge star on primal 1-forms presented in [Ayoub et al 2021](https://www.aimsciences.org/article/doi/10.3934/cpaa.2021062).

### DIFF
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -1035,19 +1035,19 @@ vector field.
   lie_derivative_flat(Val{n}, s, args...; kw...)
 
 function lie_derivative_flat(::Type{Val{0}}, s::AbstractACSet,
-                             X♭::AbstractVector, α::AbstractVector)
-  interior_product_flat(1, s, X♭, dual_derivative(0, s, α))
+                             X♭::AbstractVector, α::AbstractVector; kw...)
+  interior_product_flat(1, s, X♭, dual_derivative(0, s, α); kw...)
 end
 
 function lie_derivative_flat(::Type{Val{1}}, s::AbstractACSet,
-                             X♭::AbstractVector, α::AbstractVector)
-  interior_product_flat(2, s, X♭, dual_derivative(1, s, α)) +
-    dual_derivative(0, s, interior_product_flat(1, s, X♭, α))
+                             X♭::AbstractVector, α::AbstractVector; kw...)
+  interior_product_flat(2, s, X♭, dual_derivative(1, s, α); kw...) +
+    dual_derivative(0, s, interior_product_flat(1, s, X♭, α; kw...))
 end
 
 function lie_derivative_flat(::Type{Val{2}}, s::AbstractACSet,
-                             X♭::AbstractVector, α::AbstractVector)
-  dual_derivative(1, s, interior_product_flat(2, s, X♭, α))
+                             X♭::AbstractVector, α::AbstractVector; kw...)
+  dual_derivative(1, s, interior_product_flat(2, s, X♭, α; kw...))
 end
 
 # Euclidean geometry
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -15,14 +15,14 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
   OrientedDeltaDualComplex2D, EmbeddedDeltaDualComplex2D,
   SimplexCenter, Barycenter, Circumcenter, Incenter, geometric_center,
   subsimplices, primal_vertex, elementary_duals, dual_boundary, dual_derivative,
-  ⋆, hodge_star, δ, codifferential, ∇², laplace_beltrami, Δ, laplace_de_rham,
+  ⋆, hodge_star, inv_hodge_star, δ, codifferential, ∇², laplace_beltrami, Δ, laplace_de_rham,
   ♭, flat, ♯, sharp, ∧, wedge_product, interior_product, interior_product_flat,
   ℒ, lie_derivative, lie_derivative_flat,
   vertex_center, edge_center, triangle_center, dual_triangle_vertices,
   dual_point, dual_volume, subdivide_duals!, DiagonalHodge, GeometricHodge
 
 import Base: ndims
-using LinearAlgebra: Diagonal, dot, norm, cross, cholesky
+using LinearAlgebra: Diagonal, dot, norm, cross
 using SparseArrays
 using StaticArrays: @SVector, SVector
 
@@ -672,7 +672,15 @@ end
 ⋆(::Type{Val{n}}, s::AbstractACSet, ::DiagonalHodge) where n =
   Diagonal([ hodge_diag(Val{n},s,x) for x in simplices(n,s) ])
 
-crossdot(v1, v2) = norm(cross(v1, v2)) * sign(last(cross(v1, v2)))
+# Note that this cross product defines the positive direction for flux to
+# always be in the positive z direction. This will likely not generalize to
+# arbitrary meshes embedded in 3D space, and so will need to be revisited.
+# Potentially this orientation can be provided by the simplicial triangle
+# orientation?
+crossdot(v1, v2) = begin
+  v1v2 = cross(v1, v2)
+  norm(v1v2) * (last(v1v2) == 0 ? 1.0 : sign(last(v1v2)))
+end
 
 """ Hodge star operator from primal 1-forms to dual 1-forms.
 
@@ -780,7 +788,7 @@ function inv_hodge_star(::Type{Val{1}}, s::AbstractDeltaDualComplex2D,
 end
 function inv_hodge_star(::Type{Val{1}}, s::AbstractDeltaDualComplex2D,
                         form::AbstractVector, ::GeometricHodge)
-  inv(Matrix(⋆(Val{1}, s, GeometricHodge()))) * form
+  Matrix(⋆(Val{1}, s, GeometricHodge())) \ form
 end
 
 inv_hodge_star(::Type{Val{0}}, s::AbstractDeltaDualComplex2D, ::GeometricHodge) =
@@ -809,8 +817,8 @@ inv_hodge_star(::Type{Val{n}}, s::AbstractDeltaDualComplex1D,
 @inline δ(n::Int, s::AbstractACSet, args...; kw...) =
   δ(Val{n}, s, args...; kw...)
 @inline δ(::Type{Val{n}}, s::AbstractACSet; hodge=GeometricHodge(),
-          matrix::Type=SparseMatrixCSC{Float64}) where n =
-  δ(Val{n}, s, hodge, matrix)
+          matrix_type::Type=SparseMatrixCSC{Float64}) where n =
+  δ(Val{n}, s, hodge, matrix_type)
 @inline δ(::Type{Val{n}}, s::AbstractACSet, form::AbstractVector;
           hodge=GeometricHodge()) where n =
   δ(Val{n}, s, hodge, form)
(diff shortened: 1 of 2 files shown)


bosonbaas:
There are still a few things that need to be decided before merging this PR, specifically:

- [ ] How do we want to facilitate the user choosing between the diagonal hodge star (if using just circumcentric sub) and this one?
- [ ] Since this does not have a `hodge_diag` function, the `δ` function will need to be updated to support this implementation. (related to the first question)
- [ ] Tests need to be added to show consistency between the diagonal and new hodge star for circumcentric subdivision.

epatters:
Thanks for implementing this improved Hodge star. Re: your first question, I think we should use the dispatch-on-algorithm design pattern, used elsewhere in this package to select to the subdivision method. We should make this new Hodge star the default, but keep the diagonal Hodge star around for benchmarking and debugging purposes.

src/DiscreteExteriorCalculus.jl
```diff
@@ -664,6 +664,52 @@ end
...
+  vals = Dict{Tuple{Int64, Int64}, Float64}()
+  for t in triangles(s)
+    e = reverse(triangle_edges(s, t))
+    ev = point(s, tgt(s, e)) .- point(s, src(s,e))
+
+    dv = fill(dual_point(s, triangle_center(s, t)),3) .- dual_point(s, edge_center(s, e))
+    dv[2] *= -1
+
+    diag_dot = map(1:3) do i
+             dot(ev[i], dv[i]) / norm(ev[i])^2
```
jpfairbanks, line 693: using `ev[i]'ev[i]` will save a sqrt on every call.

src/DiscreteExteriorCalculus.jl
```diff
@@ -664,6 +664,52 @@ end
...
+meshes generated under other subdivision schemes (e.g. barycentric).
+"""
+function ⋆(::Type{Val{1}}, s::AbstractACSet)
+
+  vals = Dict{Tuple{Int64, Int64}, Float64}()
+  for t in triangles(s)
+    e = reverse(triangle_edges(s, t))
+    ev = point(s, tgt(s, e)) .- point(s, src(s,e))
+
+    dv = fill(dual_point(s, triangle_center(s, t)),3) .- dual_point(s, edge_center(s, e))
```
jpfairbanks, line 689: I think that you can avoid this fill because of the semantics of broadcasting `.-` either you can avoid it or pull it outside the subtraction.

src/DiscreteExteriorCalculus.jl
```diff
@@ -664,6 +664,52 @@ end
 ⋆(::Type{Val{n}}, s::AbstractACSet) where n =
   Diagonal([ hodge_diag(Val{n},s,x) for x in simplices(n,s) ])
 
+add_val!(d, k, v) = if k ∈ keys(d)
```
jpfairbanks, line 667: the sparse function defaults to adding values for items with repeated keys. So there is no need to do this deduping in the dictionary if this is a helper for `sparse` as used below. 

```julia

  sparse(I, J, V,[ m, n, combine])

  Create a sparse matrix S of dimensions m x n such that S[I[k], J[k]] = V[k]. The combine function is used to combine duplicates. If m and n are not specified, they are set to maximum(I) and maximum(J) respectively. If the combine function is not supplied, combine defaults to + unless the elements of V are Booleans in which case
  combine defaults to |. All elements of I must satisfy 1 <= I[k] <= m, and all elements of J must satisfy 1 <= J[k] <= n. Numerical zeros in (I, J, V) are retained as structural nonzeros; to drop numerical zeros, use dropzeros!.
```

jpfairbanks (commented):
I noticed some performance issues that I think we can easily clean up.

src/DiscreteExteriorCalculus.jl
bosonbaas: Yeah, it'd be nice to remove that fill, but I'm not sure how to get it to broadcast the first argument (which is a 3-element vector) over the second argument (which is a 3-element vector of 3-element vectors). In this last commit I've changed it out with a `map` call, but I'd be interested in learning about other solutions to this.

src/DiscreteExteriorCalculus.jl
bosonbaas: That's really cool! Definitely cleans up the implementation, thanks

bosonbaas:
I've tested the new hodge on some physical systems, and it seems to be performing as expected! I've included tests which should evaluate both the diagonal and geometric hodge stars, and added a keyword system for specifying hodge star and matrix type from any of the methods that depend on these.

I'll just be adding some tests for codecov, but the body of the code is ready for another review!

bosonbaas:
100% coverage now!

test/DiscreteExteriorCalculus.jl
```diff
@@ -6,6 +6,7 @@ using SparseArrays, StaticArrays
 
 using Catlab.CategoricalAlgebra.CSets
 using CombinatorialSpaces
+using CombinatorialSpaces.DiscreteExteriorCalculus: inv_hodge_star
```
epatters, line 9: I guess we should go ahead and export this function.

test/DiscreteExteriorCalculus.jl
```diff
@@ -150,13 +164,43 @@ subdivide_duals!(s, Barycenter())
...
+@test inv_hodge_star(1, s, hodge=DiagonalHodge()) ≈ Diagonal([-6/√5, -6, -6/√5])
+@test inv_hodge_star(1, s, [0.5, 2.0, 0.5], hodge=DiagonalHodge()) ≈ [-3/√5, -12.0, -3/√5]
 @test ⋆(s, VForm([1,2,3]))::DualForm{2} ≈ DualForm{2}([1/6, 1/3, 1/2])
-@test isapprox(δ(1,s), [ 2.236  0  2.236;
-                        -2.236  1  0;
-                         0     -1 -2.236], atol=1e-3)
+@test isapprox(δ(1,s; hodge=DiagonalHodge()), [ 2.236  0  2.236;
+                                              -2.236  1  0;
+                                               0     -1 -2.236], atol=1e-3)
+@test isapprox(δ(1,s), [ 3.0  0  3.0;
```
epatters, line 181: The specific numbers I included for the diagonal Hodge came from the paper cited in the comments above. Did you get the numbers for the geometric Hodge from somewhere? Also, it's interesting that they are round numbers up to, I assume, floating point error. Were you able to just calculate these by hand?

src/DiscreteExteriorCalculus.jl
```diff
@@ -722,14 +853,14 @@ This linear operator on primal ``n``-forms defined by ``∇² α := -δ d α``,
...
+  SimplexForm{n}(∇²(Val{n}, s, x.data; kw...))
+@inline ∇²(n::Int, s::AbstractACSet, args...; kw...) = ∇²(Val{n}, s, args...; kw...)
 
-∇²(::Type{Val{n}}, s::AbstractACSet, form::AbstractVector) where n =
-  -δ(Val{n+1}, s, d(Val{n}, s, form))
-∇²(::Type{Val{n}}, s::AbstractACSet, Mat::Type=SparseMatrixCSC{Float64}) where n =
-  -δ(Val{n+1}, s, Mat) * d(Val{n}, s, Mat)
+∇²(::Type{Val{n}}, s::AbstractACSet, form::AbstractVector; kw...) where n =
+  -δ(n+1, s, d(Val{n}, s, form); kw...)
+∇²(::Type{Val{n}}, s::AbstractACSet; matrix::Type=SparseMatrixCSC{Float64}, kw...) where n =
```
epatters, line 862: Not a big deal, but IMO, the keyword argument you are now calling `matrix` should be called `matrix_type` or `Matrix` to indicate that it is supposed to be a matrix type rather than a matrix itself.

src/DiscreteExteriorCalculus.jl
```diff
@@ -674,25 +749,73 @@ Confusingly, this is *not* the operator inverse of the Hodge star [`⋆`](@ref)
...
+  end
+end
+
+function inv_hodge_star(::Type{Val{1}}, s::AbstractDeltaDualComplex2D,
+                        ::GeometricHodge)
+  inv(Matrix(⋆(Val{1}, s, GeometricHodge())))
+end
+function inv_hodge_star(::Type{Val{1}}, s::AbstractDeltaDualComplex2D,
+                        form::AbstractVector, ::GeometricHodge)
+  inv(Matrix(⋆(Val{1}, s, GeometricHodge()))) * form
```
epatters, line 783: We want to do a linear solve rather than a matrix inverse here, right? IIRC, @jpfairbanks had some other thoughts related to this as well.

src/DiscreteExteriorCalculus.jl
```diff
@@ -19,10 +19,10 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
   ♭, flat, ♯, sharp, ∧, wedge_product, interior_product, interior_product_flat,
   ℒ, lie_derivative, lie_derivative_flat,
   vertex_center, edge_center, triangle_center, dual_triangle_vertices,
-  dual_point, dual_volume, subdivide_duals!
+  dual_point, dual_volume, subdivide_duals!, DiagonalHodge, GeometricHodge
 
 import Base: ndims
-using LinearAlgebra: Diagonal, dot, norm
+using LinearAlgebra: Diagonal, dot, norm, cross, cholesky
```
epatters, line 25: Looks like `cholesky` is not being used here.

src/DiscreteExteriorCalculus.jl
```diff
@@ -655,15 +659,86 @@ end
...
+@inline ⋆(::Type{Val{n}}, s::AbstractACSet, form::AbstractVector;
+          hodge=GeometricHodge()) where n = ⋆(Val{n}, s, form, hodge)
+
+⋆(::Type{Val{n}}, s::AbstractACSet, form::AbstractVector, ::DiagonalHodge) where n =
   applydiag(form) do x, a; a * hodge_diag(Val{n},s,x) end
-⋆(::Type{Val{n}}, s::AbstractACSet) where n =
+⋆(::Type{Val{n}}, s::AbstractACSet, ::DiagonalHodge) where n =
   Diagonal([ hodge_diag(Val{n},s,x) for x in simplices(n,s) ])
 
+crossdot(v1, v2) = norm(cross(v1, v2)) * sign(last(cross(v1, v2)))
```
epatters, line 675: Minor optimization: compute the cross product once, then use its value twice.

epatters (commented):
Thanks Andrew, this will be great to have.

test/DiscreteExteriorCalculus.jl
bosonbaas: Ah, got it! Line 169 is consistent with the values on page 13 in the paper, but for the rest I was just using the numbers generated by the operator (putting them in the tests to show consistency in future implementations, but not as evaluations of this current implementation unfortunately).
I'm afraid that I'd be more likely to err in my hand-calculation, and I'd be following the same algorithm. Would it make sense for me to put a comment around tests that don't have a direct literature reference, and to put the reference for tests that have literature support?

test/DiscreteExteriorCalculus.jl
bosonbaas: Would it make sense to also include a function for ⋆ that accepts a DualForm and calls `inv_hodge_star`?

src/DiscreteExteriorCalculus.jl
bosonbaas: Yeah, that's right. I'll put it as a basic linear solve for now, we might consider using other iterative methods here as well.

src/DiscreteExteriorCalculus.jl
bosonbaas: I have now changed these to `matrix_type`

test/DiscreteExteriorCalculus.jl
epatters: I'd prefer to introduce a new operator `⋆⁻¹`  as in the exterior calculus GAT.

bosonbaas: Ah, yeah, that makes sense.

test/DiscreteExteriorCalculus.jl
bosonbaas: Are the tests you had for the coboundary (173 in the old, 222 in the new) verified from a specific document? If not, I might just put a global "All tests below this line are not externally verified" warning after the first few verified tests.

epatters (approved):

epatters:
Thanks Andrew! I will go ahead and merge this now.

test/DiscreteExteriorCalculus.jl
epatters: Created issue #43 for this.

Issue #37, opened 2021-08-05
Lie Derivative for vector field acting on non 1-forms


bosonbaas:
I've run into the issue where the Lie derivative fails if called on a flattened vector field acting on a dual form that isn't a 1-form. The error seems to be from the `dual_derivative` and `interior_product` being called with dimension n < 0 or n > N.

Code which reproduces this error is given below, along with the stack trace:
```julia
using CombinatorialSpaces
using SparseArrays, StaticArrays

const Point2D = SVector{2,Float64}
primal_s = EmbeddedDeltaSet2D{Bool,Point2D}()
add_vertices!(primal_s, 4, point=[Point2D(-1,+1), Point2D(+1,+1),
                                  Point2D(+1,-1), Point2D(-1,-1)])
glue_triangle!(primal_s, 1, 2, 3, tri_orientation=true)
glue_triangle!(primal_s, 1, 3, 4, tri_orientation=true)
primal_s[:edge_orientation] = true
s = EmbeddedDeltaDualComplex2D{Bool,Float64,Point2D}(primal_s)
subdivide_duals!(s, Barycenter())

X♭, α = EForm([1.5, 2, 2.5, 3, 3.5]), DualForm{0}([3, 7])
ℒ(s, X♭, α)

X♭, α = EForm([1.5, 2, 2.5, 3, 3.5]), DualForm{2}([3, 7, 10, 11])
ℒ(s, X♭, α)
```
Stack trace (with CSet output abbreviated):
```
ERROR: MethodError: no method matching nsimplices(::Type{Val{3}}, ::Catlab.CSetDataStructures, NamedTuple{(:D_∂e0, :D_∂e1, :D_∂e2, :D_∂v0, :D_∂v1, :src, :tgt, :∂e0, :∂e1, :∂e2), NTuple{10, Vector{Vector{Int64}}}}})                                                            
Closest candidates are:                                                                                                                                                                                                            
  nsimplices(::Type{Val{0}}, ::Any) at ~/test/CombinatorialSpaces.jl/src/SimplicialSets.jl:56                                                                  
  nsimplices(::Type{Val{1}}, ::Any) at ~/test/CombinatorialSpaces.jl/src/SimplicialSets.jl:57                                                                  
  nsimplices(::Type{Val{2}}, ::Any) at ~/test/CombinatorialSpaces.jl/src/SimplicialSets.jl:167                                                                 
  ...                                                                                                                                                                                                                              
[...]

PR #38, opened 2021-08-05
Lie derivative for dual 0,2-forms

bosonbaas:
This PR addresses Issue #37 by creating three Lie Derivative functions for each of the dual forms.

Would it make more sense to add edge-case functions for the `dual_derivative`, `nsimplices`, and other functions that might receive dimension values out of range? This might reduce the amount of code redundancy here.


epatters (approved):
Thanks for the fix! Supporting the edge cases in the interior product and exterior derivative would be nice, but I'm also fine with this approach of specializing the Lie derivative methods. FWIW, I use the same approach in the exterior calculus GATs since they are specialized to specific dimensions.

Issue #41, opened 2021-09-08
Stackoverflow from recursion in orientation function


bosonbaas:
For large meshes, the `orient!` function causes a stack overflow error. This Issue is just a stand-in while I investigate these errors a little further, but I think it relates to the recursion, and a non-recursive implementation might prove useful.

https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/main/src/SimplicialSets.jl#L478

bosonbaas:
I have confirmed that the iterative solution presented in #42 fixes the stack overflow error I encountered.

PR #42, opened 2021-09-09
Iterative Implementation for orient_component!

bosonbaas:
This PR addresses Issue #41 by changing the `orient_component!` function from recursive to iterative. This should generate equivalent orientations of meshes.

### DIFF
--- a/src/SimplicialSets.jl
+++ b/src/SimplicialSets.jl
@@ -475,25 +475,28 @@ function orient_component!(s::AbstractACSet, x::Simplex{n},
                            x_orientation::Orientation) where {n, Orientation}
   orientations = repeat(Union{Orientation,Nothing}[nothing], nsimplices(n, s))
 
-  function orient_neighbors!(x, target)
+  orient_stack = Vector{Pair{Int64, Orientation}}()
+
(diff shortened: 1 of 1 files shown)


src/SimplicialSets.jl
```diff
@@ -475,25 +475,28 @@ function orient_component!(s::AbstractACSet, x::Simplex{n},
...
         for y in coface(n, j, s, ∂(n, i, s, x))
-          y == x || orient_neighbors!(y, next) || return false
+          y == x || push!(orient_stack, y=>next)
         end
       end
-      true
     else
       # If already visited, check that current and target orientations agree.
-      current == target
+      is_orientable = is_orientable && current == target
```
epatters, line 496: In non-orientable case, better to break from the loop, so replace the `else` clause with:

```julia
elseif current != target
  is_orientable = false
  break
end
```

epatters (commented):
Thanks Andrew! One small comment.

src/SimplicialSets.jl
bosonbaas: Gotcha, that definitely works! Makes sense to not do extra work in this function if it has already failed.

PR #44, opened 2021-09-20
Flexible Wedge Product for 0-forms

bosonbaas:
This PR adds an option for different wedge product weighting schemes. 

This implementation is based on Griebel et. al 2017 (Upwind Schemes for Scalar Advection-Dominated Problems in the Discrete Exterior Calculus). Through this, we can reproduce several techniques currently used in finite element/finite volume/finite difference methods, like central differences and upwinding.

### DIFF
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -957,30 +957,31 @@ the discrete primal-primal wedge product introduced in (Hirani, 2003, Chapter 7)
 and (Desbrun et al 2005, Section 8). It depends on the geometric embedding and
 requires the dual complex.
 """
-∧(s::AbstractACSet, α::SimplexForm{k}, β::SimplexForm{l}) where {k,l} =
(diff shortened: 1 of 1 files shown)


jpfairbanks:
@lukem12345 has this been subsumed by other PRs?

lukem12345:
No, in the mean time we have produced wedge products between e.g. primal and dual forms. But methods that perform upwinding have not been implemented.

jpfairbanks:
Now that we are doing more fluid stuff, upwinding seems really helpful. I've labeled this good first issue to get over the finish line.

jpfairbanks:
Now that we have a general approach to directional derivatives (using Lie(x,y) as our version of d/dx y), could we use that to do upwinding? We specify a vector field x to be the direction in which we define as "upwind" and then reweight operators like wedge or lie derivatives by scalar multiplication by weights?

lukem12345:
Certainly the 1 dimensional case will be easy to implement and test as a preliminary PR.

PR #49, opened 2021-12-09
Refactor type hierarchy for delta sets

epatters:
I've been intending to make this change, enabled by the struct acsets refactor, for some time. It creates a type hierarchy for delta sets of different dimensions, where, for example, one-dimensional delta sets (`AbstractDeltaSet1D`) are distinguished from delta sets having simplices of dimension at least one (`HasDeltaSet1D`). By tightening the types, this PR paves the way for implementing DEC with cubical sets but supporting the same generic functions.

Although none of the interfaces have changed, this PR should probably be considered breaking because it changes the acset schemas for delta sets in the first dimension (`src` and `tgt` are renamed with a scheme consistent with higher dimensions).

### DIFF
--- a/src/SimplicialSets.jl
+++ b/src/SimplicialSets.jl
@@ -7,14 +7,15 @@ maps but not the degeneracy maps of a simplicial set. In the future we may add
 support for simplicial sets. The analogy to keep in mind is that graphs are to
 semi-simpicial sets as reflexive graphs are to simplicial sets.
 
-Also provided are the fundamental operators on simplicial sets used in virtually
+Also provided are the fundamental operators on simplicial sets used in nearly
 all geometric applications, namely the boundary and coboundary (discrete
-exterior derivative). For additional operators, see the
+exterior derivative) operators. For additional operators, see the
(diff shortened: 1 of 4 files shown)


src/SimplicialSets.jl
```diff
@@ -30,50 +31,98 @@ using LinearAlgebra: det
...
 
-Delta sets in 1D are the same as graphs, and this type is just an alias for
-`Graph`. The face maps [`∂`](@ref) translates the graph-theoretic terminology
-into simplicial terminology.
+Delta sets in 1D are isomorphic to graphs (in the category theorist's sense).
+The source and target of an edge can be accessed using the face maps [`∂`](@ref)
+(simplicial terminology) or `src` and `tgt` maps (graph-theoretic terminology).
+More generally, this type implements the graphs interface in `Catlab.Graphs`.
 """
-const DeltaSet1D = Graph
```
epatters, line 54: @jpfairbanks, RIP to your favorite line in CombinatorialSpaces.

PR #56, opened 2023-02-03
Set orientations when making dual simplices

lukem12345:
Using recent versions of Catlab, the MeshInterop tests are failing due to edge orientations being set to `nothing` instead of a default value, as was the previous behavior.

This pull request solves this issue by calling `orient!` inside the `make_dual_simplices` functions if orientations are not set. If a valid orientation is not found, then arbitrary orientations are set.

Currently, all `CombinatorialSpaces.jl` tests pass, as well as relevant tests in `Decapodes.jl` that depend on this package.


jpfairbanks:
Seems like a bit of a foot gun to default to invalid orientations instead of raising an exception. Let's get the build working with this approach, but open an issue about how to handle it in general with @epatters.

PR #57, opened 2023-06-07
Support 3D Simplicial Sets

lukem12345:
This PR is to track progress on implementing 3D (semi-)simplicial sets in SimplicialSets.jl.

Currently, CombinatorialSpaces.jl supports only 1D and 2D simplicial sets. Although some methods are implemented for arbitrary dimension, such as `orient!`, these methods cannot take advantage of this flexibility due to the lack of higher dimensional simplicial complexes stored as ACSets.

Although there may be some benefits to refactoring SimplicialSets.jl to support arbitrary ND, the diminishing utility for dimensions beyond 3 or 4 for physical applications justifies focusing on a hand-written implementation of 3D simplicial sets for the time being. This is not to say that there is no utility to such an approach. One advantage is that of long-term maintainability at the expense of a larger up-front engineering effort.

### DIFF
--- a/src/SimplicialSets.jl
+++ b/src/SimplicialSets.jl
@@ -1,3 +1,11 @@
+# TODO: Document where appropriate where the compositional numbering vs.
+# simplicial identity numbering scheme is in use in this API.
+
+# TODO: Add a function that checks whether a complex is_manifold_like. (See
+# Hirani's definition.)
+
+# TODO: Would an accessor for the boundary edges of a tetrahedron be useful?
+
 """ Simplicial sets in one, two, and three dimensions.
 
 For the time being, this module provides data structures only for [delta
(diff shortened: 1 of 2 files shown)


src/SimplicialSets.jl
```diff
@@ -221,6 +230,35 @@ This is the shape of the binary composition operation in a category.
...
+  #map(x -> only(x[:Tri].func), homs)
+  # This is more elegant, exploiting the explicit representation of our axioms
+  # from the schema, instead of implicitly exploiting them like below.
+  # This method requires us to provide the schema for s as well, so we may have
+  # to refactor around multiple dispatch, or always data-migrate to
+  # SchDeltaSet2D.
+
+  # TODO: Is there a better idiom for `isempty(x) && return []`?
+  e₀s = coface(1,0,s,v₂) ∩ coface(1,1,s,v₁)
+  isempty(e₀s) && return []
```
lukem12345, line 254: Don't return Vector{Any}

test/SimplicialSets.jl
```diff
@@ -39,11 +39,14 @@ add_sorted_edges!(s, [2,4], [3,3])
 # 1D oriented simplicial sets
 #----------------------------
 
+using Test
```
lukem12345, line 42: Delete

test/SimplicialSets.jl
```diff
@@ -156,6 +162,216 @@ add_vertices!(s, 3, point=[Point3D(1,0,0), Point3D(0,1,0), Point3D(0,0,1)])
...
+@test is_semi_simplicial(s, 2)
+@test is_semi_simplicial(s, 3)
+@test tetrahedron_vertices(s, 1) == [1,2,3,5]
+@test tetrahedron_vertices(s, 2) == [2,3,4,5]
+@test sort(unique(reduce(vcat, edge_vertices(s, tetrahedron_edges(s,1))))) ==
+  tetrahedron_vertices(s,1)
+@test sort(unique(reduce(vcat, edge_vertices(s, tetrahedron_edges(s,2))))) ==
+  tetrahedron_vertices(s,2)
+
+# Six tetrahedra forming a cube.
```
lukem12345, line 205: Hellper function

PR #59, opened 2023-07-01
Matrix sharps, matrix flats, subdivision, and link

lukem12345:
This PR implements subdivision as a data migration, matrix versions of the sharp and flat operators, an alternative weighting scheme for the sharp operator, and the star, closed star, and link operators from S68 of Munkres' Elementary Algebraic Topology.

_Some of these changes may need to get shifted to a different branch_. In particular, subdivision cannot yet automatically compute barycenters of simplices until the `attrMigr` branch of Catlab is merged. Relatedly, refactoring the DualComplex schemas as a data migration that "remembers" e.g. which triangle a supporting dual triangle came from also depends on this functionality.

The star, closed star, and link operators were implemented as higher-level helper functions which will become more useful as 3D operators are becoming supported. This (not-Hodge) star essentially recursively computes cofaces, which is a common pattern in this codebase. e.g. grabbing all tetrahedra around a vertex v is simply star(v).

While developing the matrix version of the sharp operator, it was noted that the sharp of an EForm generated by a constant form would generate a primal vector field with the correct direction, but of incorrect magnitude. In particular, vectors with few neighbors would be too small, and those with many neighbors would be too large. By replacing the weighting denominator in Eq. 5.8.1 from Hirani with | star v | instead of | sigma n |, we successfully recover the original field (up to floating point error). This alternate discrete sharp operator can be selected via dispatching with `AltPPSharp`. Further, a sharp operator from Desbrun can be selected via `DesbrunSharp`.

As implemented at time of writing, the matrix-flat multiplied by a vector X of `SVector`s will return a vector of length 1 `SArray`s. When X is wrapped in `DualVectorField()`, this result is automatically `reinterpret`ed to a vector of floats. This fix is a little unusual, but I am not sure how to structure the flat matrix to avoid this.

PPS
PPS_ALT


lukem12345:
Due to scheduling, this PR consists of a few disparate features:
- Subdivision of non-embedded 1D delta sets w/o point attributes
- Star, Closed Star, and Link operators from Munkres. a.k.a. "topological helper functions"
- Matrix version of sharp
- Matrix version of flat
- Alternate weighting scheme for flat

All these features have tests.

When considering which features to merge, we should note:
- Only 1D subdivision-as-data-migration is supported. Due to outstanding [Catlab PR #823](https://github.com/AlgebraicJulia/Catlab.jl/pull/823), barycentric and other automatic weighting of the `point` attribute is blocked. Although the current `subdivide` functions pass tests, perhaps this feature should live on a different branch for the time being.
- Star, Closed Star, and Link are implemented for vertex input only. I believe that the current implementation is helpful as-is, however.
- See my previous comment for the return type of the matrix version of flat
- The Desbrun weighting scheme is rather poor, and I don't imagine it would be used in practice. However, it may still be nice to offer this for completeness' sake.
- Likewise, I believe that the Hirani weighting scheme is inferior to the "Alternate" weighting scheme due to behavior mentioned in my previous comment. It may be best to change the default weighting scheme to "Alternate."

lukem12345:
Todo:
- [] Use QR decomposition in the new sharp operator

jpfairbanks:
I think you are probably better off with GMRES or PCG because the QR decomp will be dense.

lukem12345:
Oh, the QR decomp is just for the tiny (3×3 in R3, 3×2 in R2) matrices. For
each triangle T, I split this tiny matrix into its rows, and place each row
in the corresponding position in the (num-triangles × num-edges) sparse
matrix.

Maybe QR decomposition is not what I want here,  since the application of
the entire operation is still just sparse matrix multiplication.

src/FastDEC.jl
```diff
@@ -230,6 +234,150 @@ function dec_wedge_product(::Type{Tuple{1,1}}, sd::HasDeltaSet2D)
...
+"""
+    function wedge_pd_01_mat(sd::HasDeltaSet)
+
+Returns a matrix that can be multiplied to a primal 0-form, before being
+elementwise-multiplied by a dual 1-form, encoding the wedge product.
+
+This function assumes barycentric means and performs bilinear interpolation. It
+is not known if this definition has appeared in the literature or any code.
+"""
+function wedge_pd_01_mat(sd::HasDeltaSet)
```
olynch, line 288: This could be generalized by taking as an argument the type of matrix to create. Also, we could have a type `WedgeProduct` which wrapped `sd`. When applied to arguments, this would dynamically compute the wedge product. But then this function could be written as something like:

`concretize(w::WedgeProduct, A::Type{<:AbstractMatrix})::A`

Then we can reuse `concretize` for all sorts of similar problems. This is beyond the scope of this PR though: I'll make an issue for this.

src/Meshes.jl
```diff
@@ -284,4 +292,72 @@ function makeSphere(minLat, maxLat, dLat, minLong, maxLong, dLong, radius)
   return s, north_pole_idx, south_pole_idx
 end
 
-end
\ No newline at end of file
+"""    function tri_345()
```
olynch, line 295: what is a 3-4-5 triangle: can you add a link to this docstring?

src/Meshes.jl
```diff
@@ -284,4 +292,72 @@ function makeSphere(minLat, maxLat, dLat, minLong, maxLong, dLong, radius)
...
+  primal_s = EmbeddedDeltaSet2D{Bool,Point3D}()
+  add_vertices!(primal_s, 3, point=[Point3D(0,0,0), Point3D(3,0,0), Point3D(3,4,0)])
+  glue_triangle!(primal_s, 1, 2, 3, tri_orientation=true)
+  primal_s[:edge_orientation] = true
+  s = EmbeddedDeltaDualComplex2D{Bool,Float64,Point3D}(primal_s)
+  subdivide_duals!(s, Barycenter())
+  (primal_s, s)
+end
+
+"""    function fri_345()
```
olynch, line 309: Why is this called `fri`?

src/Meshes.jl
```diff
@@ -284,4 +292,72 @@ function makeSphere(minLat, maxLat, dLat, minLong, maxLong, dLong, radius)
...
+  glue_sorted_triangle!(primal_s, 9, 6, 8)
+  primal_s[:edge_orientation] = true
+  s = EmbeddedDeltaDualComplex2D{Bool,Float64,Point3D}(primal_s)
+  subdivide_duals!(s, Barycenter())
+  (primal_s, s)
+end
+
+"""    function right_unit_hypot()
+
+Return the primal and dual mesh of a right triangle with unit hypotenuse.
```
olynch, line 349: This does not uniquely specify the triangle. Do you mean an isosceles right triangle?

src/SimplicialSets.jl
```diff
@@ -905,4 +906,93 @@ function nonboundaries(s::HasDeltaSet, ::Type{Simplex{n}}) where n
...
+
+link(s::AbstractDeltaSet1D, v::Int) = link(s, v, E)
+link(s::AbstractDeltaSet2D, v::Int) = link(s, v, Tri)
+link(s::AbstractDeltaSet3D, v::Int) = link(s, v, Tet)
+
+""" Link of a vertex in a delta set.
+
+Munkres §2 ≈ "The set St̄(v) - St(v)."
+
+Return a vector of simplices of dimensions 0 to n.
```
olynch, line 981: This docstring was confusing: you mean returns a vector where the element at index `i` is a set of simplices of dimension `i - 1`. This should probably be a tuple rather than a vector, and you should wrap it with a struct with a good name, like

```
struct Simplices{Ts <: Tuple}
  by_dimension::Ts
end
```

src/FastDEC.jl
olynch: :green_circle:

src/Meshes.jl
olynch: :red_circle:

src/Meshes.jl
olynch: :red_circle:

src/Meshes.jl
olynch: :red_circle:

src/SimplicialSets.jl
olynch: :green_circle:

olynch (changes requested):
I've left some comments. The comments with :green_circle: should be addressed in a future PR, the comments with :red_circle: should be addressed in this PR.

PR #64, opened 2023-11-07
Added FastDEC

GeorgeR227:
Some fundamental DEC operators that have been optimized for the DeltaSet ACSets are added in their own separate file. The reasons why these have been added into their own file is as follows:

1. The new DEC operators will probably not work or they may return bad results when the DualDeltaSet is somehow modified by beyond the subdivide_duals! method. However, there is no user restriction on modifying these ACSets so these operators are not "safe". This was the reason for the "fast_" prefix, implying that some assumptions are made to allow optimizations.

2. These new DEC operators sometimes return matrices and sometimes don't, specifically the Geometric Inverse Hodge for 1-Forms and the different Wedge Products. Since older implementations of more complicated operators, like Laplace de-Rham, only expect matrices this means that the old implementations cannot simply be replaced.

Separate from these concerns is that there is a desire to benchmark these operators but there are no large meshes stored with this package. Instead, these kinds of meshes are currently supported by the Decapodes package. So either mesh generation code must be upstreamed from Decapodes or the benchmarking will be done in that package, which is a strange division of responsibility.


Project.toml
```diff
@@ -6,9 +6,12 @@ version = "0.6.0"
 
 [deps]
 ACSets = "227ef7b5-1206-438b-ac65-934d6da304b8"
+Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
```
lukem12345, line 9: Artifacts `compat`?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+using SparseArrays: sparse
+using LinearAlgebra
+using Base.Iterators
+using ACSets.DenseACSets: attrtype_type
+using Catlab, Catlab.CategoricalAlgebra.CSets
+using ..SimplicialSets, ..DiscreteExteriorCalculus
+
+export dec_boundary, dec_differential, dec_dual_derivative, dec_hodge_star, dec_inv_hodge_star, dec_wedge_product
+
+# TODO: This relies on the assumption of a well ordering of the
```
lukem12345, line 13: I think some of these TODOs can be turned into XXXs, and maybe should be given in docstrings as well.

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+using Base.Iterators
+using ACSets.DenseACSets: attrtype_type
+using Catlab, Catlab.CategoricalAlgebra.CSets
+using ..SimplicialSets, ..DiscreteExteriorCalculus
+
+export dec_boundary, dec_differential, dec_dual_derivative, dec_hodge_star, dec_inv_hodge_star, dec_wedge_product
+
+# TODO: This relies on the assumption of a well ordering of the 
+# the dual space simplices. If changed, use dec_p_wedge_product_zero
+function dec_p_wedge_product(::Type{Tuple{0,1}}, sd; float_type=Float64)
```
lukem12345, line 15: Type annotation for `sd`?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+        coeffs[1, i] = (d_areas[i] + d_areas[i+shift]) / area
+        coeffs[2, i] = (d_areas[i+2*shift] + d_areas[i+3*shift]) / area
+        coeffs[3, i] = (d_areas[i+4*shift] + d_areas[i+5*shift]) / area
+    end
+    # TODO: This is assuming that meshes don't have too many entries
+    # This type should be settable by the user and default set to Int32
+    # e = Array{Int64}(undef, 3, ntriangles(sd))
+    e = Array{Int32}(undef, 3, ntriangles(sd))
+
+    e[1, :] = ∂(2, 0, sd)
```
lukem12345, line 107: Is there a FastDEC implementation of \partial that could be used here?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+    if (negate)
+        V .= -1 .* V
+    end
+
+    (I, J, V)
+end
+
+# Diagonal Hodges 
+
+# TODO: Check this Hodge with a 1D mesh
```
lukem12345, line 275: This could be a test

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+function dec_p_hodge_diag(::Type{Val{0}}, sd::AbstractDeltaDualComplex1D; float_type=Float64)
+    num_v_sd = nv(sd)
+
+    hodge_diag_0 = zeros(float_type, num_v_sd)
+
+    v1_list = @view sd[:D_∂v1]
+    dual_lengths = @view sd[:dual_length]
+
+    for d_edge_idx in eachindex(v1_list)
+        v1 = v1_list[d_edge_idx]
```
lukem12345, line 285: Is this an `enumerate` pattern?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+
+function dec_p_hodge_diag(::Type{Val{0}}, sd::AbstractDeltaDualComplex2D; float_type=Float64)
+    hodge_diag_0 = zeros(float_type, nv(sd))
+
+    dual_edges_1 = @view sd[:D_∂e1]
+    dual_v_1 = @view sd[:D_∂v1]
+    dual_areas = @view sd[:dual_area]
+
+    for dual_tri in eachindex(dual_edges_1)
+        v = dual_v_1[dual_edges_1[dual_tri]]
```
lukem12345, line 308: Likewise an `enumerate` here?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+    J = Vector{Int32}()
+    V = Vector{float_type}()
+
+    edge_centers = @view sd[:edge_center]
+    tri_centers = @view sd[:tri_center]
+
+    # points::Vector{Point3{Float64}} = sd[:point]
+    # dual_points::Vector{Point3{Float64}} = sd[:dual_point]
+
+    #TODO: Figure out how to type these since both Point2D and Point3D can be used
```
lukem12345, line 377: Can you pass the point type in as a type parameter of `sd`?

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+
+dec_inv_hodge_star(::Type{Val{1}}, sd::AbstractDeltaDualComplex1D, ::GeometricHodge; float_type=Float64) =
+    dec_inv_hodge_star(Val{1}, sd, DiagonalHodge(), float_type=float_type)
+
+dec_inv_hodge_star(::Type{Val{0}}, sd::AbstractDeltaDualComplex2D, ::GeometricHodge; float_type=Float64) =
+    dec_inv_hodge_star(Val{0}, sd, DiagonalHodge(), float_type=float_type)
+
+function dec_inv_hodge_star(::Type{Val{1}}, sd::AbstractDeltaDualComplex2D, ::GeometricHodge; float_type=Float64)
+    hdg_lu = LinearAlgebra.factorize(-1 * dec_hodge_star(1, sd, GeometricHodge(), float_type=float_type))
+    x -> hdg_lu \ x
```
lukem12345, line 466: This "caches" `hdg_lu` right?

src/Meshes.jl
```diff
@@ -0,0 +1,306 @@
...
+    end
+    if (connect_long)
+      glue_sorted_triangle!(s, south_pole_idx, south_parallel_start,
+                            south_parallel_start+num_meridians-1)
+    end
+  end
+  return s, north_pole_idx, south_pole_idx
+end
+
+## Tests
```
lukem12345, line 287: These can be moved to `test` somewhere.

src/FastDEC.jl
GeorgeR227: Yes, this will compute the proper factorization and then save it for later use. The only detail to note is that this should eventually switch to an iterative solver which may be better for large but sparse systems.

src/FastDEC.jl
GeorgeR227: This is actually already done on line 99 of test/Operators.jl, I just forgot to remove the TODO.

src/FastDEC.jl
GeorgeR227: The partial that's being used here is really just a wrapper for the ```face``` method which just grabs some subparts of the ACSet so it's already pretty lightweight and direct. Maybe there is some way to trick this out to use the boundary operator but the generation feels fast enough here anyway.

src/FastDEC.jl
GeorgeR227: Yes, fixed now

src/FastDEC.jl
GeorgeR227: Yes, also fixed now

src/FastDEC.jl
GeorgeR227: Annotated for HasDeltaSet

src/CombinatorialSpaces.jl
```diff
@@ -10,9 +10,14 @@ include("SimplicialSets.jl")
 include("DiscreteExteriorCalculus.jl")
 include("MeshInterop.jl")
```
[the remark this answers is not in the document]
GeorgeR227: Removed now

src/FastDEC.jl
GeorgeR227: I've added docstrings for the wedge products but I should also add them for the rest of the operators.

Project.toml
GeorgeR227: Added compat for 1.9, same as Julia.

src/FastDEC.jl
```diff
@@ -0,0 +1,472 @@
...
+
+dec_hodge_star(::Type{Val{2}}, sd::AbstractDeltaDualComplex2D, ::GeometricHodge; float_type=Float64) =
+    dec_hodge_star(Val{2}, sd, DiagonalHodge(), float_type=float_type)
+
+crossdot(v1, v2) = begin
+    v1v2 = cross(v1, v2)
+    norm(v1v2) * (last(v1v2) == 0 ? 1.0 : sign(last(v1v2)))
+end
+
+function dec_hodge_star(::Type{Val{1}}, sd::AbstractDeltaDualComplex2D, ::GeometricHodge; float_type::DataType = Float64)
```
lukem12345, line 365: Since the Hodge star operator depends on the metric, it is inappropriate for this function to be called on any delta dual complex object, (such as `OrientedDeltaDualComplex2D`), which does not have `Point` information. (I think the theory technically does not require explicit points to require a metric, but our practical implementation depends on such.)

So, it is okay IMO for `sd` to be explicitly  `::EmbeddedDeltaDualComplex2D{_, _, point}` for this operator.

src/FastDEC.jl
GeorgeR227: I've edited the hodges to have their ```sd``` types be EmbeddedDeltaDualSets now and below are the results. The before is on the left, where we were only checking for AbstractDeltaDualComplexes and the right is the new with the stricter Embedded type. These benchmarks are ran on an Ico5 mesh.

image
image
image

There are clear improvements to making the types stricter and we should check to see if any other operators can benefit from this

PR #71, opened 2024-03-01
Faster Dual Mesh Generation

GeorgeR227:
This pull requests improves some of the dual mesh generation code to allow it to generate dual mesh information faster and with a lower memory footprint. This dual information is generated with the same functions but now a FastMesh() is passed in as an argument to dispatch to the new code.

This code is separate from the original mesh generation code as this new code bypasses a lot of the ACSet functionality to achieve better performance. A better fix would offer improvements to the underlying ACSet accessing and setting.


GeorgeR227:
image

Some benchmarks run on functions going from the primal mesh all the way to a fully filled out dual mesh.

jpfairbanks (commented):
It looks like there are some places where you have an extra layer of dispatch (see line comments). Can we eliminate that layer? 

Is the reason to bypass the ACSet machinery the need to put all the `@inbounds`es and `@view`s around? Can you open an upstream issue on ACSets and work with @olynch to fix that? 

It looks like you aren't using `incident` anywhere here. Generally when you have a graph algorithm, if you can write it in terms of `edges` instead of `for all v; for n in neighbors(v)` you can get better cache performance and better parallelism. Is that what is happening here? 

Can you benchmark the two versions as part of the docs?

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,341 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+
+function subdivide_duals!(sd::EmbeddedDeltaDualComplex1D, gen::FastMesh, args...)
+  subdivide_duals_1d!(sd, gen, args...)
+  precompute_volumes_1d!(sd, gen)
+end
+
+function subdivide_duals!(sd::EmbeddedDeltaDualComplex2D, gen::FastMesh, args...)
+  subdivide_duals_2d!(sd, gen, args...)
+  precompute_volumes_2d!(sd, gen)
+end
```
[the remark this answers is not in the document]
GeorgeR227: By method do you just mean that these functionality is part of the class itself and not just floating around? If that's the case then I agree, I notice there is a lot of passing around of the mesh. On this I was just following the structure of the original code for consistency.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,341 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+struct PointType{T}
+  point_type::T
+end
+
+point_wrapper(sd::EmbeddedDeltaDualComplex1D{_o, _l, point_type} where {_o, _l}) where point_type = return PointType{point_type}
+point_wrapper(sd::EmbeddedDeltaDualComplex2D{_o, _l, point_type} where {_o, _l}) where point_type = return PointType{point_type}
+
+subdivide_duals_1d!(sd::EmbeddedDeltaDualComplex1D, fm::FastMesh, alg) = subdivide_duals_1d!(sd, fm, point_wrapper(sd), alg)
+subdivide_duals_1d!(sd::EmbeddedDeltaDualComplex2D, fm::FastMesh, alg) = subdivide_duals_1d!(sd, fm, point_wrapper(sd), alg)
+subdivide_duals_2d!(sd::EmbeddedDeltaDualComplex2D, fm::FastMesh, alg) = subdivide_duals_2d!(sd, fm, point_wrapper(sd), alg)
```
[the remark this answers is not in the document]
GeorgeR227: The problem here is that the with the ```subdivide_duals``` and the ```precompute_volumes```, I need that point information for good performance. However, the code is structured in a way that when ```subdivide_duals_2d!``` is called, the first thing it does is run ```subdivide_duals_1d!``` and then goes on to the 2D specific code. 

The subdivide code matches for ```HasDeltaSet2D``` and ```HasDeltaSet1D```. This is because of the type hierarchy, if you have ```HasDeltaSet2D``` then you also have ```HasDeltaSet1D``` so the dispatch works. But there is no hierarchy like that with ```EmbeddedDeltaDualComplex2D``` and ```EmbeddedDeltaDualComplex1D```, so if you have the 2D you are not seen as having the 1D.

Here I kinda hack my way into by leaving the ```HasDeltaSet2D``` and ```HasDeltaSet1D``` so dispatch works but since I know the mesh is really always an ```EmbeddedDeltaDualComplex```, I just grab the point type with the wrapper and pass it down.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,341 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+
+precompute_volumes_1d!(sd::HasDeltaSet1D, fm::FastMesh) = precompute_volumes_1d!(sd, fm, point_wrapper(sd))
+precompute_volumes_2d!(sd::HasDeltaSet2D, fm::FastMesh) = precompute_volumes_2d!(sd, fm, point_wrapper(sd))
```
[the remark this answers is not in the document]
GeorgeR227: Yes this is done mainly for performance, I do this to pass down the point information which is crucial because I am using MVectors (mutable static vectors) of point_types to remove heap allocations. However, I don't expect a user should ever call the version of the function expecting the point_type, so this automatically picks it up for them.

GeorgeR227:
As for the dispatch, there might be a cleaner way of writing this code out but this current version works and I fear that cleaner versions would need deeper changes to the CombinatorialSpaces package, specifically to way the type hierarchy is structured.

The main reason for bypassing the ACSet machinery, like with the fast operators, is that using that machinery always costs heap allocations which in turn slows down code significantly and balloons the memory footprint, as shown in the benchmarks above. I will definitely make an issue on ACSets about improving this but for now I personally feel that I've deviated from the GPU Decapodes for long enough that I should return to that first.

Coincidentally, a lot of the FastDEC operators I wrote are improved based on that same idea, so flowing down tri->edge->vertex instead of climbing up which would require ```incident``` calls. However, this dual mesh code is essentially one-to-one from the original, just with more views and avoidance of indexing the ACSet directly to reduce allocations.

I can for sure add a benchmark of the two different generations.

src/DiscreteExteriorCalculus.jl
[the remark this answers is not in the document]
jpfairbanks: I had forgotten that subdivide 2D needs to first subdivide 1D. Now I understand why you are doing it this way. We should probably be using value types for dimension 0,1,2,3 everywhere in the codebase, but that is a larger project than this PR.

src/DiscreteExteriorCalculus.jl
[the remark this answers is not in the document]
jpfairbanks: It seems like you should still be able to remove the extra wrapper of PointType and just dispatch on MVector.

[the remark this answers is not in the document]
GeorgeR227: Are you essentially saying instead of having ```::Type{PointType{point_type}}```, to just have ```::Type{point_type}``` instead?

jpfairbanks:
That is great, I really like what you just wrote about flowing down Tri -> Edge -> Vertex being faster. Can you add a doc page noting that with your benchmarks. You can write something like a Performance Tips page and this can be a first tip. It is always better to avoid incident calls by working form the top dimensional simplicies down to vertices rather than working bottom up. 

If we can actually purge enough of the operators code of incident, then we could actually disable the morphism indexes, which would really improve the memory performance. Any calls to incident are going to be a locality problem which is a cache/parallel/GPU performance trap. Definitely an issue worth opening for follow up.

GeorgeR227:
Yeah, that Performance Tips page sounds like a good idea to disseminate what I've learned about writing performant ACSet code. On terms of performance I completely agree with you about ```incident``` in that we should never rely on this kind of brute lookup of data.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+
+function (::Type{S})(t::AbstractDeltaSet2D, gen::FastMesh) where S <: AbstractDeltaDualComplex2D
+  s = S()
+  copy_primal_2D!(s, t)
+  make_dual_simplices_2d!(s, gen)
+  return s
+end
+
+copy_primal_1D!(t::EmbeddedDeltaDualComplex1D{_o, _l, point_type} where {_o, _l}, s) where point_type = copy_primal_1D!(t, s, point_type)
+copy_primal_1D!(t::EmbeddedDeltaDualComplex2D{_o, _l, point_type} where {_o, _l}, s) where point_type = copy_primal_1D!(t, s, point_type)
```
jpfairbanks, line 873: is this some kind of julia witchcraft? Why do we need these extra methods? If for performance, that should be commented so that we know if it is safe to remove in future versions.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+  src_v0 = @view s[:∂v0]
+  src_v1 = @view s[:∂v1]
+
+  tgt_v0 = @view t[:∂v0]
+  tgt_v1 = @view t[:∂v1]
+
+  for (i, e) in enumerate(e_range)
+    tgt_v0[e] = src_v0[i]
+    tgt_v1[e] = src_v1[i]
+  end
```
jpfairbanks, line 895: Can we do this with broadcasting? making a view and then iterating over it seems unnecessary. I guess you could get slightly better cache performance by doing the copy in this order. but a broadcasting assignment should make a copy.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+    for (i, tri) in enumerate(tri_range)
+      tgt_tri_orient[tri] = src_tri_orient[i]
+    end
+  end
+
+end
+
+make_dual_simplices_1d!(s::AbstractDeltaDualComplex1D, gen::FastMesh) = make_dual_simplices_1d!(s, E, gen)
+
+function make_dual_simplices_1d!(sd::HasDeltaSet1D, ::Type{Simplex{n}}, gen::FastMesh) where n
```
jpfairbanks, line 939: you never us `n` in this code. Why are you dispatching on the type of the Simplex, when you have 1d in the function name? It seems like we should either use the dimension in the function name, or dispatch on dimension with an argument

jpfairbanks (changes requested):

src/DiscreteExteriorCalculus.jl
lukem12345: `n` is used here to distinguish the degree of the simplicial complex. If `n` is equal to the current degree (e.g. if we are orienting edges on a 1D complex), then we must orient the edges. Otherwise (e.g. if we are orienting edges on 2D complex), we can choose arbitrary orientation.

src/DiscreteExteriorCalculus.jl
GeorgeR227: The main reason is the need for the point type. Getting this type is crucial for good performance and can only be gotten from the ```EmbeddedDeltaDualComplex```. However, the main function checks for ```HasDeltaSet1D``` so that both 1D and 2D embedded meshes can use that function.

src/DiscreteExteriorCalculus.jl
GeorgeR227: The views are necessary due to the extraneous allocations caused by ACSet indexing. I could broadcast over the view but I vaguely remember that not being as good for some reason.

src/DiscreteExteriorCalculus.jl
jpfairbanks: Ok so add a comment like:

> This nested dispatch is necessary for performance as of Julia 1.10 because we want to dispatch on a concrete value for point_type, but use an abstract type for `t::HasDeltaSet1D`. Future improvements to dispatch performance could render this unnecessary.

src/DiscreteExteriorCalculus.jl
jpfairbanks: Ugh, we need to fix that upstream.

src/DiscreteExteriorCalculus.jl
GeorgeR227: The ```n``` is used in the orientation code starting on line 970 and is actually checked on line 976. The reason is that 2D meshes will use this function but only the triangle orientations seem to matter and so the orientation code here shouldn't be run. On line 996 you can see the same function declared but expecting a ```AbstractDeltaDualComplex2D``` and feeding it a ```Tri``` instead.

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+    d_e2[D_triangles[i]] .= view(D_edges01[ev+1], triangle_edges[e+1])
+  end
+
+  if has_subpart(sd, :tri_orientation)
+    # If orientations are not set, then set them here.
+    if any(isnothing, sd[:tri_orientation])
+      # 2-simplices only need to be orientable if the delta set is 2D.
+      # (The 2-simplices in a 3D delta set need not represent a valid 2-Manifold.)
+      if n == 2
+        orient!(sd, Tri) || error("The 2-simplices of the given 2D delta set are non-orientable.")
```
jpfairbanks, line 1074: missing test coverage here is suspicious because I would expect this to get hit on every call.

src/DiscreteExteriorCalculus.jl
GeorgeR227: Yeah that sounds good I'll add a comment like that.

jpfairbanks:
@epatters, a lot of this code is working around a lack of high performance views in the ACSets code. Is this another consequence of adding variable attribute support?

Do we have the software engineering resources to dedicate to getting this fixed upstream?

src/DiscreteExteriorCalculus.jl
GeorgeR227: Yeah you were right, all of the 2D meshes we use have pre-oriented triangles. I included a simple triangle mesh in there, with no orientation, and tests still pass.

GeorgeR227:
Backing up the need for this, here are the latest benchmarks. These are run on a triangulated grid with 1,502,001 dual vertices, 4,502,000 dual edges and 3,000,000 dual triangles.

image

GeorgeR227:
Here are the functions used as well: 

image

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
...
+      else
+        sd[findall(isnothing, sd[:edge_orientation]), :edge_orientation] = zero(attrtype_type(sd, :Orientation))
+      end
+    end
+
+    # TODO: Need to use for-loop here since orientations are by default "nothing"
+    # Need to find a way around this
+    edge_orient = @view sd[:edge_orientation]
+    d_edge_orient = @view sd[:D_edge_orientation]
+    for i in edges(sd)
```
lukem12345, line 987: I see that `@inbounds` is not used here, despite you using `@inbounds` for similar loops

src/DiscreteExteriorCalculus.jl
```diff
@@ -851,6 +852,400 @@ function precompute_volumes_2d!(s::HasDeltaSet2D)
   end
 end
 
+### Begin faster dual mesh generation ###
```
lukem12345, line 855: What is the rationale for including this code here, not in `FastDEC`?

lukem12345:
> Backing up the need for this, here are the latest benchmarks. These are run on a triangulated grid with 1,502,001 dual vertices, 4,502,000 dual edges and 3,000,000 dual triangles.

These should go in `src/Benchmarks.jl`, since we host similar comparisons there.

src/DiscreteExteriorCalculus.jl
GeorgeR227: Yeah, adding an ```@inbounds``` would be fine.

src/DiscreteExteriorCalculus.jl
GeorgeR227: This code was meant to overload the already existing functions in this file, namely ```subdivide duals!``` and the constructor for dual meshes. Because of the way I introduced this code, we could very well move it into a different file.

GeorgeR227: It might be better to move it into its own file ```FastDECMesh``` or something since ```FastDEC``` is already large and this is separate functionality from DEC operator code.

GeorgeR227:
This is based off of the release of ACSets v0.2.19.

Updated fast ```subdivide_duals!``` benchmarks. These will be pulled into the original ACSets implementation:

```julia
julia> GC.gc(true)

julia> @time subdivide_duals!(sd, Barycenter())
  2.834950 seconds (30.39 M allocations: 2.828 GiB, 6.22% gc time)

julia> GC.gc(true)

julia> @time subdivide_duals!(sd, FastMesh(), Barycenter())
  0.044771 seconds (4 allocations: 240 bytes)
```

Benchmarks for ```EmbeddedDeltaDualComplex2D```. The fast code has not been adapted but this is meant to show that performance of the original is nearing the fast code. As a result the original code will probably remain as is since the fast code is not as maintainable and not as superior anymore.

```julia
julia> GC.gc(true)

julia> @time EmbeddedDeltaDualComplex2D{Bool, Float64, Point2D}(s);
  0.536737 seconds (5.45 M allocations: 556.682 MiB, 10.63% gc time)

julia> @time EmbeddedDeltaDualComplex2D{Bool, Float64, Point2D}(s, FastMesh());
  0.244258 seconds (3.48 M allocations: 321.850 MiB)
```

GeorgeR227:
Comparing with the main CombinatorialSpaces branch with ACSets 0.2.17 active.
```julia
[ Info: Generated Primal Mesh
[ Info: Original Dual Mesh Generation
  8.935594 seconds (81.95 M allocations: 12.278 GiB, 11.48% gc time)
```

src/DiscreteExteriorCalculus.jl
```diff
@@ -233,11 +233,37 @@ negatenz((I, V)) = (I, negate.(V))
...
 """ Construct 1D dual complex from 1D delta set.
 """
-function (::Type{S})(t::AbstractDeltaSet1D) where S <: AbstractDeltaDualComplex1D
-  s = S()
-  copy_parts!(s, t)
-  make_dual_simplices_1d!(s)
-  return s
+function (::Type{S})(s::AbstractDeltaSet1D) where S <: AbstractDeltaDualComplex1D
+  t = S()
+  copy_primal_1D!(t, s) # TODO: Revert to copy_parts! when performance is improved
```
lukem12345, line 238: Can you add an issue to ACSets.jl that you are having to circumvent `copy_parts!`

src/DiscreteExteriorCalculus.jl
```diff
@@ -364,30 +390,71 @@ Supports different methods of subdivision through the choice of geometric
...
+
+  @inbounds for v in vertices(sd)
+    sd[v, :dual_point] = sd[v, :point]
   end
-  for e in edges(s)
-    s[edge_center(s,e), :dual_point] = geometric_center(
-      point(s, edge_vertices(s, e)), alg)
+
+  @inbounds for e in edges(sd)
+    point_arr[1] = sd[sd[e, :∂v0], :point]
```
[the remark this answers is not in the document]
lukem12345: And elsewhere that pattern is used in this pr. So we can easily come back when ACSets is updated

src/DiscreteExteriorCalculus.jl
```diff
@@ -364,30 +390,71 @@ Supports different methods of subdivision through the choice of geometric
...
+
+  @inbounds for e in parts(sd, :DualE)
+    point_arr[1] = sd[sd[e, :D_∂v0], :dual_point]
+    point_arr[2] = sd[sd[e, :D_∂v1], :dual_point]
+
+    sd[e, :dual_length] = volume(point_arr)
   end
 end
 
+# function subdivide_duals!(s::EmbeddedDeltaDualComplex1D, args...)
```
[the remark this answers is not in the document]
GeorgeR227: Done.

src/DiscreteExteriorCalculus.jl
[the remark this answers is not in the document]
GeorgeR227: I'll place a TODO above every function that does this.

src/DiscreteExteriorCalculus.jl
GeorgeR227: Yes, I'll get one out once I've looked into the issue some more. I'm guessing this again stems from vectorized getting/setting plus other underlying issues.

src/DiscreteExteriorCalculus.jl
```diff
@@ -832,29 +919,81 @@ function ∧(::Type{Tuple{1,1}}, s::HasDeltaSet2D, α, β, x::Int)
...
+  @inbounds for t in parts(sd, :DualTri)
+    point_arr[1] = sd[sd[sd[t, :D_∂e1], :D_∂v1], :dual_point]
+    point_arr[2] = sd[sd[sd[t, :D_∂e2], :D_∂v0], :dual_point]
+    point_arr[3] = sd[sd[sd[t, :D_∂e0], :D_∂v0], :dual_point]
+
+    sd[t, :dual_area] = volume(point_arr)
   end
 end
 
+# function subdivide_duals!(s::EmbeddedDeltaDualComplex2D, args...)
```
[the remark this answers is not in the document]
GeorgeR227: Done.

test/Benchmarks.jl
```diff
@@ -154,4 +154,47 @@ for op in sort(collect(keys(results)))
     println("----------------------------------------------------------------")
 end
 
+@info "Beginning Dual Mesh Generation Benchmarks"
+begin 
+    mesh_size = 100
+    grid_spacings = [1.0, 0.8, 0.5, 0.4, 0.25, 0.2]
+    point_type = Point2D
+    benchmark_dual_meshes = map(gs -> triangulated_grid(mesh_size, mesh_size, gs, gs, point_type), grid_spacings);
```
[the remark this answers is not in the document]
GeorgeR227: What do you mean by on demand?

test/Benchmarks.jl
```diff
@@ -125,25 +125,25 @@ begin
...
+dec_op_results = run(dec_op_suite, verbose = true, seconds = 1)
 
-results = run(suite, verbose = true, seconds = 1)
-
-for op in sort(collect(keys(results)))
-    test = median(results[op])
+for op in sort(collect(keys(dec_op_results)))
+    test = median(dec_op_results[op])
 
     println("Operator: $op")
```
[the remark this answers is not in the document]
GeorgeR227: I thought about it but I'm not sure about how to abstract it. Mainly due to lots of formatting things changing between suites, like output text, order of output, etc.

src/DiscreteExteriorCalculus.jl
```diff
@@ -364,30 +390,71 @@ Supports different methods of subdivision through the choice of geometric
...
-  precompute_volumes_1d!(s)
+function subdivide_duals!(sd::EmbeddedDeltaDualComplex1D{_o, _l, point_type} where {_o, _l}, alg) where point_type
+  subdivide_duals_1d!(sd, point_type, alg)
+  precompute_volumes_1d!(sd, point_type)
 end
 
-function subdivide_duals_1d!(s::HasDeltaSet1D, alg)
-  for v in vertices(s)
-    s[vertex_center(s,v), :dual_point] = point(s, v)
+# TODO: Write helpers for getting the correct mesh data
```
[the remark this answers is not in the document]
GeorgeR227: I'm checking to see if using the built-in CombinatorialSpaces.jl data retrievers, like ```triangle_vertices```, is just as performant.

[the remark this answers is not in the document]
GeorgeR227: Seems like it. I've added my own helper since ```dual_edge_vertices``` didn't seem to exist along with a test. This also cleans up the functions so tupling there is moot.

test/Benchmarks.jl
```diff
@@ -154,4 +154,47 @@ for op in sort(collect(keys(results)))
...
+end
+
+dual_mesh_suite = BenchmarkGroup()
+
+begin
+    dual_mesh_suite["Barycenter"] = BenchmarkGroup()
+    dual_mesh_suite["Circumcenter"] = BenchmarkGroup()
+
+    for (i, grid_spacing) in enumerate(grid_spacings)
+        dual_mesh_suite["Barycenter"]["Grid Spacing: $(grid_spacing)"] = @benchmarkable create_dual_mesh($(benchmark_dual_meshes[i]), $point_type, $(Barycenter())) gcsample=true seconds=60
```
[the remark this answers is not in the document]
GeorgeR227: Surprisingly you can but it just converts the whole thing into one big string:
```julia
2-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "Barycenter" => 1-element BenchmarkTools.BenchmarkGroup:
          tags: []
          "(grid_spacings = [1.0, 0.8, 0.5, 0.4, 0.25, 0.2],)" => Benchmark(evals=1, seconds=5.0, samples=10000)
```

src/DiscreteExteriorCalculus.jl
```diff
@@ -945,55 +936,36 @@ function precompute_volumes_2d!(sd::HasDeltaSet2D, p::Type{point_type}) where po
...
+# TODO: Replace the individual accesses with vector accesses
 function set_dual_volumes_2d!(::Type{Val{2}}, sd::HasDeltaSet2D, ::Type{point_type}) where point_type
 
   point_arr = MVector{3, point_type}(undef)
 
   @inbounds for t in parts(sd, :DualTri)
-    point_arr[1] = sd[sd[sd[t, :D_∂e1], :D_∂v1], :dual_point]
-    point_arr[2] = sd[sd[sd[t, :D_∂e2], :D_∂v0], :dual_point]
-    point_arr[3] = sd[sd[sd[t, :D_∂e0], :D_∂v0], :dual_point]
+    p1, p2, p3 = dual_triangle_vertices(sd, t)
```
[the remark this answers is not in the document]
GeorgeR227: It returns an```SVector``` to be precise, which would be better than an Array due to no allocations. However, using vectorized access is significantly worse since memory usage will scale with mesh size while now it is totally constant. Essentially you'll be killing performance gains for some aesthetics.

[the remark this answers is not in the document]
GeorgeR227: Once vectorized access also drops to 0 alloc, then this solution seems good to me.

GeorgeR227:
This is still part of https://github.com/AlgebraicJulia/ACSets.jl/issues/135. The issue's been closed since the fix is good enough but it is not totally solved.

test/Benchmarks.jl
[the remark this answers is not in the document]
GeorgeR227: These handles aren't used once, they're used by both ```Barycenter``` and ```Circumcenter``` benchmark groups. The only advantage would be that all meshes don't have to be kept in memory all the time but I think by interpolating the meshes into the benchmarks this is the case anyway.

lukem12345 (approved):

Issue #72, opened 2024-03-05
Infinite recursion calling loadmesh with a non AbstractMeshKey


quffaro:
```
dset = DeltaSet1D()
add_vertices!(dset, 1)
dset |> loadmesh
```
This code will produce an error
```
Stacktrace:
     [1] loadmesh(s::DeltaSet1D, subdivision::Circumcenter)
       @ CombinatorialSpaces.Meshes ~/.julia/packages/CombinatorialSpaces/Wg7wx/src/Meshes.jl:74--- the last 1 lines are repeated 79982 more times ---
```

This occurs because loadmesh will call `subdivide_duals!(loadmesh(s), subdivision)`, and iterate.

I imagine there should be some way to eyeball whether this will recurse. I'm willing to assign this to myself.

lukem12345:
Hi, which version of `loadmesh` is this calling? Can you please do `@which`? This function is not meant to be used this way.

quffaro:
It's calling it here:
```
loadmesh(s)
     @ CombinatorialSpaces.Meshes ~/.julia/packages/CombinatorialSpaces/Wg7wx/src/Meshes.jl:74
```

lukem12345:
Ok. It seems that there are two issues here. The typing on `loadmesh` should not allow this usage.

quffaro:
Is it possible to distinguish at the type level "bad" objects from "good" ones so we could just overload `loadmesh`, or should there be code which checks for recursive badness (and throws an error?)?

jpfairbanks:
It is this fallback method [here](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/c8def89a5664003f390494fcbf5ffd5c5c9f7c55/src/Meshes.jl#L74).

This should have a generic type like `AbstractMeshIdentifier` that has concrete subtypes for all our existing meshes so that the fallback method isn't defined for `Any`.

lukem12345:
Yes this should subtype `AbstractMeshKey`. The name "load" was originally chosen because we were literally loading a mesh from a zip file (i.e. Artifacts). What Matt wants is a helper function that takes a primal mesh, creates a dual mesh, and subdivides it.

lukem12345:
[...]

PR #73, opened 2024-03-06
Remove bad loadmesh function

lukem12345:


lukem12345:
The bad definition of `loadmesh` does not appear to be called in neither this repo nor in Decapodes. So I've gone ahead and deleted it. On merging this PR, I will add another issue about deleting the `loadmesh` function entirely, and just further overloading `EmbeddedDeltaDualComplex2D` or `EmbeddedDeltaSet2D` for the appropriate mesh keys.

lukem12345:
Close #72

GeorgeR227 (approved):
If this isn't used anywhere then this is fine. I also agree with removing loadmesh in general since some of the newer meshes we have, like triangulated_grid, no longer rely on it. We should instead have a single function that handles all kinds of primal mesh loading for the user.

PR #77, opened 2024-05-01
Remove unnecessary extra wrapping of valtypes

KevinDCarlson:
There's no difference in functionality between `f(::Type{Val{n}}) where n =...` and `f(::Val{n}) where n = ...`, but the latter is both cleaner and the intended usage: you plug in the unique *instance* of the type `Val{n}` rather than plugging in the type. I'm not able to find any difference in performance between the two approaches but I wanted to clean this up anyway.

### DIFF
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -164,14 +164,14 @@ vertex_center(s::HasDeltaSet, args...) = s[args..., :vertex_center]
 """
 edge_center(s::HasDeltaSet1D, args...) = s[args..., :edge_center]
 
-subsimplices(::Type{Val{1}}, s::HasDeltaSet1D, e::Int) =
+subsimplices(::Val{1}, s::HasDeltaSet1D, e::Int) =
   SVector{2}(incident(s, edge_center(s, e), :D_∂v0))
 
-primal_vertex(::Type{Val{1}}, s::HasDeltaSet1D, e...) = s[e..., :D_∂v1]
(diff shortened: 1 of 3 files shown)


epatters:
Have you done benchmarks? I'm not sure how it shakes out this case, but my understanding is that there *is* sometimes a reason to prefer `::Type{Val{T}}` to `::Val{T}`, namely that `Val{t}`, where `t` is a literal, is a constant available at compile time and can therefore be constant propagated, whereas `Val(t)` is value that is only constructed and dispatched at runtime.

At least at one time this was a meaningful distinction that could be detected in certain situations. Perhaps, due to compiler improvements, it is no longer relevant. I would suggest getting advice from someone who actually understands how the Julia compiler works.

jpfairbanks:
This PR would make things much more idiomatic, but we do need the benchmarks, since this code is so critical to decapodes performance. A quick and dirty way to check, since you are touching such a large fraction of the codebase is to just run the tests before and after the change. If the tests take longer, then likely this is the culprit.

KevinDCarlson:
> This PR would make things much more idiomatic, but we do need the benchmarks, since this code is so critical to decapodes performance. A quick and dirty way to check, since you are touching such a large fraction of the codebase is to just run the tests before and after the change. If the tests take longer, then likely this is the culprit.

Oh I definitely did that, it gets like 1-2% faster IIRC. Also tried some synthetic tests but wasn’t able to show the difference ever affects performance.

lukem12345:
Thanks for the catch @KevinDCarlson . After adding in the Galerkin-mode multigrid mode, the type inference stuff became more bottlenecky. So changes are made in that PR.

Linking: PR #196

Issue #76, opened 2024-04-26
Parameterize the Hodge Formulation for Fast DEC operators


lukem12345:
Certain operators such as the interior product between dual 1-forms "hard-code" the choice of GeometricHodge.

Although the interpolation method that this interior product relies on assumes perfect barycentric interpolation, this operator can use the DiagonalHodge formulation. (With the universal caveat that it will be more accurate on "well-structured" meshes.)

These operators should parameterize the choice of Hodge star:
- [ ] Dual Laplacian
- [ ] Interior Product between Dual 1-forms
- [ ] Lie derivative between Dual 1-forms

lukem12345:
The typo in [this doc-string](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/ca54d0c13367289bab17f73d2dd5f0d8a98210a1/src/FastDEC.jl#L310) should be amended.

GeorgeR227:
For dec_wedge_product_pd and dec_wedge_product_dp instead of calling ```only.()``` in the cached function, this should be applied on the flat-sharp matrix beforehand, meaning that we only do matrix multiplication when computing.

lukem12345:
Older operators such as the reference Laplacian implementation should also support such parameterization.

PR #78, opened 2024-05-01
Move sign inside wedge pd, and solve for pressure in euler

lukem12345:
Previously, we were multiplying by the sign of a dual 2-form after performing the primal-dual 1-1 wedge product at the outermost "test level." It makes sense to curry the sign-accommodation code inside these wedge product operators themselves.

This PR contains two further improvements to the tests: solving for pressure in the test for Euler's equations, and a standalone interior product dual-dual test set.

### DIFF
--- a/src/FastDEC.jl
+++ b/src/FastDEC.jl
@@ -15,8 +15,7 @@ export dec_boundary, dec_differential, dec_dual_derivative, dec_hodge_star, dec_
        dec_wedge_product_dd,
        Δᵈ
 
-"""
-    dec_p_wedge_product(::Type{Tuple{0,1}}, sd::EmbeddedDeltaDualComplex1D)
+"""    dec_p_wedge_product(::Type{Tuple{0,1}}, sd::EmbeddedDeltaDualComplex1D)
 
 Precomputes values for the wedge product between a 0 and 1-form.
 The values are to be fed into the wedge_terms parameter for the computational "c" varient. 
@@ -27,8 +26,7 @@ function dec_p_wedge_product(::Type{Tuple{0,1}}, sd::EmbeddedDeltaDualComplex1D)
     return (hcat(convert(Vector{Int32}, sd[:∂v0])::Vector{Int32}, convert(Vector{Int32}, sd[:∂v1])::Vector{Int32}), simplices(1, sd))
 end
 
-"""
-    dec_p_wedge_product(::Type{Tuple{0,1}}, sd::EmbeddedDeltaDualComplex2D)
+"""    dec_p_wedge_product(::Type{Tuple{0,1}}, sd::EmbeddedDeltaDualComplex2D)
 
 Precomputes values for the wedge product between a 0 and 1-form.
 The values are to be fed into the wedge_terms parameter for the computational "c" varient. 
@@ -41,8 +39,7 @@ end
 
 # XXX: This assumes that the dual vertice on an edge is always the midpoint
 # TODO: Add options to change 0.5 to a different float
-"""
-    dec_c_wedge_product!(::Type{Tuple{0,1}}, wedge_terms, f, α, val_pack)
+"""    dec_c_wedge_product!(::Type{Tuple{0,1}}, wedge_terms, f, α, val_pack)
 
 Computes the wedge product between a 0 and 1-form.
 Use the precomputational "p" varient for the wedge_terms parameter. 
@@ -59,8 +56,7 @@ function dec_c_wedge_product!(::Type{Tuple{0,1}}, wedge_terms, f, α, val_pack)
     return wedge_terms
 end
 
-"""
-    dec_p_wedge_product(::Type{Tuple{0,2}}, sd::EmbeddedDeltaDualComplex2D{Bool, float_type, _p} where _p) where float_type
+"""    dec_p_wedge_product(::Type{Tuple{0,2}}, sd::EmbeddedDeltaDualComplex2D{Bool, float_type, _p} where _p) where float_type
 
 Precomputes values for the wedge product between a 0 and 2-form.
 The values are to be fed into the wedge_terms parameter for the computational "c" varient. 
@@ -101,8 +97,7 @@ function dec_p_wedge_product(::Type{Tuple{0,2}}, sd::EmbeddedDeltaDualComplex2D{
     return (primal_vertices, coeffs, simples)
 end
 
-"""
-    dec_c_wedge_product!(::Type{Tuple{0,2}}, wedge_terms, f, α, val_pack)
+"""    dec_c_wedge_product!(::Type{Tuple{0,2}}, wedge_terms, f, α, val_pack)
 
 Computes the wedge product between a 0 and 2-form.
 Use the precomputational "p" varient for the wedge_terms parameter. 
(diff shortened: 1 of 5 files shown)


GeorgeR227:
Since you're already making changes, this PR should also address issue #76. At the very least moving the ```only.()``` call outside the returned function.

lukem12345:
@jpfairbanks Can you give this a review this afternoon or Monday? This branch is used for the current NS simulations under investigation

src/FastDEC.jl
```diff
@@ -703,15 +685,14 @@ function interior_product_dd(::Type{Tuple{1,2}}, s::SimplicialSets.HasDeltaSet)
...
-  (f,g) -> hs1 * only.(♭♯_m * ((Λ01_m * ihs0 * g) .* f))
+  (f,g) -> hs1 * ♭♯_m * ((Λ01_m * ihs0 * g) .* f)
 end
 
 """    function ℒ_dd(::Type{Tuple{1,1}}, s::SimplicialSets.HasDeltaSet)
 
 Given a dual 1-form and a dual 1-form, return their lie derivative as a dual 1-form.
 """
 function ℒ_dd(::Type{Tuple{1,1}}, s::SimplicialSets.HasDeltaSet)
-  # TODO: Check signs.
```
jpfairbanks, line 714: Signs are checked? High Confidence in an absence of sign errors.

jpfairbanks (approved):

src/FastDEC.jl
lukem12345: Yes when I wrote that TODO I was referring to the (-1)^(p*(n-p)) factors, i.e. the negative signs in the function that is returned here. This is correct

jpfairbanks: Actually, you should do `-(a + b)` instead of `-(a) - b`, right?

jpfairbanks:
@lukem12345, we had a race condition, I noticed that you have an -a - b instead of -(a+b).

src/FastDEC.jl
lukem12345: I will run some `@btime`s on large meshes and commit to main if I notice an improvement

lukem12345:
I will run some `@btime`s and report back if I find an improvement!

lukem12345:
I did not find a difference in runtime or allocs.

```julia-repl

julia> using CombinatorialSpaces, GeometryBasics, BenchmarkTools;                                                                                                   
                                                                                                                                                                    
julia> Point3D = Point3{Float64};                                                                                                                                   
                                                                                                                                                                    
julia> s = loadmesh(Icosphere(5));                                                                                                                                  
                                                                                                                                                                    
julia> sd = EmbeddedDeltaDualComplex2D{Bool, Float64, Point3D}(s);                                                                                                  
                                                                                                                                                                    
julia> subdivide_duals!(sd, Barycenter());                                                                                                                          
                                                                                                                                                                    
julia> ldd = ℒ_dd(Tuple{1,1}, sd);                                                                                                                                  
                                                                                                                                                                    
julia> function myℒ_dd(::Type{Tuple{1,1}}, s::SimplicialSets.HasDeltaSet)                                                                                           
         # ℒ := -diuv - iduv                                                                                                                                        
         d0 = dec_dual_derivative(0, s)                                                                                                                             
         d1 = dec_dual_derivative(1, s)                                                                                                                             
         i1 = interior_product_dd(Tuple{1,1}, s)                                                                                                                    
         i2 = interior_product_dd(Tuple{1,2}, s)                                                                                                                    
                                                                                                                                                                    
         (f,g) ->                                                                                                                                                   
           -(d0 * i1(f,g) +                                                                                                                                         
             i2(f,d1 * g))                                                                                                                                          
       end;                                                                                                                                                         
                                                                                                                                                                    
julia> mdd = myℒ_dd(Tuple{1,1}, sd);                                                                                                                                
                                                                                                                                                                    
julia> u = ones(ne(sd));                                                                                                                                            
                                                                                                                                                                    
julia> all(mdd(u,u) .== ldd(u,u))                                                                                                                                   
true                                                                                                                                                                
                                                                                                                                                                    
julia> @btime ldd(u,u);                                                                                                                                             
  820.285 μs (52 allocations: 808.38 KiB)                                                                                                                           
                                                                                                                                                                    
julia> @btime mdd(u,u);                                                                                                                                             
  822.990 μs (52 allocations: 808.38 KiB)
```

The types given by `@code_warntype` is similar for both, just reordered:
```julia-repl

f::Vector{Float64}                                                                                                                                                                                 
  g::Vector{Float64}                                                                                                                                                                                 
Body::Any                                                                                                                                                                                            
1 ─ %1  = Core.getfield(#self#, :d0)::SparseArrays.SparseMatrixCSC{Int8, Int32}                                                                                                                      
│   %2  = Core.getfield(#self#, :i1)::CombinatorialSpaces.FastDEC.var"#32#33"{LinearAlgebra.Diago                                                                                                    
ol, Float64, Point{3, Float64}}, SparseArrays.SparseMatrixCSC{Float64, Int64}, CombinatorialSpace                                                                                                    
astDEC.var"#30#31"{SparseArrays.UMFPACK.UmfpackLU{Float64, Int32}}}                                                                                                                                  
│   %3  = (%2)(f, g)::Any                                                                                                                                                                            
│   %4  = (%1 * %3)::Any                                                                                                                                                                             
│   %5  = Core.getfield(#self#, :i2)::CombinatorialSpaces.FastDEC.var"#34#35"{SparseArrays.Sparse                                                                                                    
C{Float64, Int32}, LinearAlgebra.Diagonal{Float64, Vector{Float64}}}                                                                                                                                 
│   %6  = Core.getfield(#self#, :d1)::SparseArrays.SparseMatrixCSC{Int8, Int32}                                                                                                                      
│   %7  = (%6 * g)::Vector{Float64}                                                                                                                                                                  
│   %8  = (%5)(f, %7)::Vector{Float64}                                                                                                                                                               
│   %9  = (%4 + %8)::Any                                                                                                                                                                             
│   %10 = -%9::Any                                                                                                                                                                                   
└──       return %10
```

lukem12345:
The `i1` function returns `Any`...

lukem12345:
Tracing back through the function calls, the `sign` function, which returns a vector of positive and negative ones corresponding to the sign of a simplex, is returning `::Any`. This is inside of the primal-dual 1,1 wedge product:

``julia-repl

Body::Any                                                                                                                                                                                            
1 ─ %1 = CombinatorialSpaces.FastDEC.:*::Core.Const(*)                                                                                                                                               
│   %2 = Core.getfield(#self#, :sd)::EmbeddedDeltaDualComplex2D{Bool, Float64, Point{3, Float64}}                                                                                                    
│   %3 = CombinatorialSpaces.FastDEC.sign(2, %2)::Any                                                                                                                                                
│   %4 = Core.getfield(#self#, :Λ_cached)::CombinatorialSpaces.FastDEC.var"#7#8"{Tuple{Matrix{Int3                                                                                                   
2}, Matrix{Float64}, UnitRange{Int64}}}                                                                                                                                                              
│   %5 = Core.getfield(#self#, :♭♯_m)::SparseArrays.SparseMatrixCSC{Float64, Int64}                                                                                                                  
│   %6 = (%5 * g)::Vector{Float64}                                                                                                                                                                   
│   %7 = (%4)(f, %6)::Vector{Float64}                                                                                                                                                                
│   %8 = Base.broadcasted(%1, %3, %7)::Any                                                                                                                                                           
│   %9 = Base.materialize(%8)::Any                                                                                                                                                                   
└──      return %9
```

lukem12345:
> Tracing back through the function calls, the `sign` function, which returns a vector of positive and negative ones corresponding to the sign of a simplex, is returning `::Any`.

This is because the type of the orientation is encoded as a type parameter to `EmbeddedDeltaDualComplex2D`.

One way to work around this is to hard-code `Bool` a la https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/ab1ea66268c631b3964f023dd2d4d41fdbdfd9ed/src/DiscreteExteriorCalculus.jl#L1101

PR #79, opened 2024-05-02
Adapted loadmesh and Artifact.toml for new Icospheres

quffaro:


Artifacts.toml
```diff
@@ -26,9 +26,16 @@ git-tree-sha1 = "7db4629f78161cdd08310b6f37ade7d50604f55d"
...
+    [[icospheres1.download]]
+		sha256 = "5f644886bc5c1bf3bde780e6f0738c89dd02eea83fe40653446ae9eaf1e18236"
+		url = "https://raw.githubusercontent.com/AlgebraicJulia/DecapodesMeshes.jl/main/meshes/icospheres1.tar.gz"
+
 [point_map]
 git-tree-sha1 = "926b99fcce23b7e01e19921f157a48a2fb3f458c"
 
     [[point_map.download]]
     sha256 = "ccdc946bc553fa5a677a3ac97eb49656bb444e27273f168bafa6cf7964b1c1f6"
-    url = "https://raw.githubusercontent.com/AlgebraicJulia/DecapodesMeshes.jl/main/meshes/point_map.tar.gz"
```
lukem12345, line 34: What happened here in the diff? Did you remove a new line?

lukem12345:
We need to test that these icospheres are loaded in correctly. You can use the pattern from 1-5 here https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/main/test/Meshes.jl#L56

PR #90, opened 2024-05-22
Use QR decomposition in DD sharp

### LINKED ISSUE #89: QR decomposition for Least-Squares in LLS sharp
In the definition of the sharp matrix for LLSDDSharp:
https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/main/src/DiscreteExteriorCalculus.jl#L781

There is a pinv(X'X) https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/729946c9d984ace99ac4e0c4724df44f96f6ccff/src/DiscreteExteriorCalculus.jl#L804

This should be replaced by a QR decomposition of Q. Ideally, we would do an implicit solve with `Rx = Q'b` but since you want to realize a matrix for this, you could at least take the reduced `inv(R)*Q'`.

lukem12345:
Close #89 

I only did a quick spot check that the `LLS` matrix computed via the `pinv` method matched the `LLS` matrix for a single triangle for a single dual 1-form. Those matrices were equivalent (save for a ~`1e-17` entry in the `pinv` `LLS` being replaced by a `0.0` entry in the `qr` `LLS`, which I took as a good sign.)

Tests run, but certain tolerance checks no longer pass, including those operators which make use of the `LLSDDSharp` internally.

I need to dive into the test outputs and verify whether it is a tolerance issue, or whether it is due to vector fields no longer satisfying the tangent condition.

### DIFF
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -28,9 +28,9 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
 import Base: ndims
 import Base: *
 import LinearAlgebra: mul!
-using LinearAlgebra: Diagonal, dot, norm, cross, pinv, qr
+using LinearAlgebra: Diagonal, dot, norm, cross, pinv, qr, ColumnNorm
 using SparseArrays
-using StaticArrays: @SVector, SVector, SMatrix
+using StaticArrays: @SVector, SVector, SMatrix, MMatrix
 using GeometryBasics: Point2, Point3
 
 const Point2D = SVector{2,Float64}
@@ -801,8 +801,13 @@ function ♯_mat(s::AbstractDeltaDualComplex2D, ::LLSDDSharp)
       end
     # TODO: Move around ' as appropriate to minimize transposing.
     X = stack(de_vecs)'
-    QRX = qr(X)
-    LLS = pinv(QRX.R) * QRX.Q'
+    # See: https://arxiv.org/abs/1102.1845
+    QRX = qr(X, ColumnNorm())
+    P = zeros(MMatrix{length(QRX.p),length(QRX.p), pt.parameters[2], length(QRX.p)*2})
+    for (i,j) in zip(1:size(P,1),QRX.p)
+      P[i,j] = one(pt.parameters[2])
+    end
+    LLS = P' * pinv(QRX.R) * QRX.Q'
     #LLS = pinv(X'*(X))*(X')
     for (i,e) in enumerate(tri_edges)
       ♯_m[t, e] = LLS[:,i]'*weights[i]
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -802,9 +802,10 @@ function ♯_mat(s::AbstractDeltaDualComplex2D, ::LLSDDSharp)
     # TODO: Move around ' as appropriate to minimize transposing.
     X = stack(de_vecs)'
     # See: https://arxiv.org/abs/1102.1845
-    QRX = qr(X, ColumnNorm())
-    LLS = (pinv(QRX.R) * QRX.Q')[QRX.p,:]
+    #QRX = qr(X, ColumnNorm())
+    #LLS = (inv(QRX.R) * QRX.Q')[QRX.p,:]
     #LLS = pinv(X'*(X))*(X')
+    LLS = pinv(X)
     for (i,e) in enumerate(tri_edges)
       ♯_m[t, e] = LLS[:,i]'*weights[i]
     end
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -28,7 +28,7 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
 import Base: ndims
 import Base: *
 import LinearAlgebra: mul!
-using LinearAlgebra: Diagonal, dot, norm, cross, pinv
+using LinearAlgebra: Diagonal, dot, norm, cross, qr
 using SparseArrays
 using StaticArrays: @SVector, SVector, SMatrix
 using GeometryBasics: Point2, Point3
@@ -801,7 +801,9 @@ function ♯_mat(s::AbstractDeltaDualComplex2D, ::LLSDDSharp)
       end
     # TODO: Move around ' as appropriate to minimize transposing.
     X = stack(de_vecs)'
-    LLS = pinv(X'*(X))*(X')
+    QRX = qr(X)
+    LLS = inv(QRX.R) * QRX.Q'
+    #LLS = pinv(X'*(X))*(X')
     for (i,e) in enumerate(tri_edges)
       ♯_m[t, e] = LLS[:,i]'*weights[i]
     end
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -28,7 +28,7 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
 import Base: ndims
 import Base: *
 import LinearAlgebra: mul!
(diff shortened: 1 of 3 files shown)


jpfairbanks (approved):

lukem12345:
Error can be introduced by realizing the inverse computed via the QR decomposition and storing it inside the sharp matrix. We can explore an option that caches the QR decomposition for each triangle, and while using a matrix with the sparsity pattern of $d_1$.

However, the interpolating musical operation `\sharp\flat` would no longer be a single matrix.

We could resolve this by supporting all options.

lukem12345:
The `LLS` matrix computed via the `qr` method is a 3x3 matrix of `NaN`s for the [tri_345 mesh](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/729946c9d984ace99ac4e0c4724df44f96f6ccff/src/Meshes.jl#L294).

`X` in this case is
```julia
julia> X
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3): 
  0.5   1.33333    0.0
 -1.0  -0.666667   0.0
 -0.5   0.666667  -0.0
```

The `LLS` computed via the `pinv` method is
```julia
julia> pinv(X'*(X))*(X')
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
 -8.63507e-17  -0.666667     -0.666667
  0.5          -5.55112e-17   0.5
  0.0           0.0           0.0
```

lukem12345:
@jpfairbanks Please advise: Do you want to use `pinv` instead of `inv` on `QRX.R`?

jpfairbanks:
I was expecting that qr produced the reduced Q and R factors. Would that avoid this possibility? I guess X could be rank deficient instead of overconstrained.

jpfairbanks:
The qr function docs say that you have to enable pivoting to get the thin QR decomposition. I think that using that will give you an R that is always full rank. 

```quote
  The following functions are available for the QR objects: inv, size, and \. When A is rectangular, \ will return a least squares solution and if the solution is not unique, the one with smallest norm is returned. When A is not full rank, factorization
  with (column) pivoting is required to obtain a minimum norm solution.

  Multiplication with respect to either full/square or non-full/square Q is allowed, i.e. both F.Q*F.R and F.Q*A are supported. A Q matrix can be converted into a regular matrix with Matrix. This operation returns the "thin" Q factor, i.e., if A is m×n with
  m>=n, then Matrix(F.Q) yields an m×n matrix with orthonormal columns. To retrieve the "full" Q factor, an m×m orthogonal matrix, use F.Q*I or collect(F.Q). If m<=n, then Matrix(F.Q) yields an m×m orthogonal matrix.
```

lukem12345:
This does not appear to work. I'm not certain what the formula is to compute a realization of a pseudoinverse matrix using the pivoted QR decomposition that can be spread out into our sharp matrix.
```julia
julia> X
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
  0.5   1.33333    0.0
 -1.0  -0.666667   0.0
 -0.5   0.666667  -0.0

julia> QRX = qr(X, ColumnNorm())
StaticArrays.QR{MMatrix{3, 3, Float64, 9}, MMatrix{3, 3, Float64, 9}, MVector{3, Int64}}([-0.816496580927726 -9.337025903963966e-18 -0.5773502691896258; 0.4082482904638631
-0.7071067811865475 -0.5773502691896257; -0.4082482904638631 -0.7071067811865475 0.5773502691896257], [-1.632993161855452 -0.6123724356957945 0.0; 0.0 1.0606601717798214 0.0; 0.0 0.0 -0.0], [2, 1, 3])

julia> QRX.R
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
 -1.63299  -0.612372   0.0
  0.0       1.06066    0.0
0.0       0.0       -0.0
                                                                                                                                                                                                                                                                                                                              julia> inv(QRX.R)
3×3 MMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
 NaN  NaN  NaN
 NaN  NaN  NaN
 NaN  NaN  NaN
```

lukem12345:
These authors provide a formula on page 5:
https://arxiv.org/abs/1102.1845

lukem12345:
Commit a6dabb4 realizes the pseudoinverse matrix according to that provided formula. I performed a quick spot-check by computing the `LLS` matrix using the method on `main`, and a method using the formula for the pseudoinverse via the pivoted QR decomposition. Observe that in the method on `main`, a particular entry is `-1.90099e-17`, and the corresponding entry computed via the QR method is `3.29719e-17`.

```julia
julia> pinv(X'*(X))*(X')
2×3 MMatrix{2, 3, Float64, 6} with indices SOneTo(2)×SOneTo(3):
 0.21  -1.90099e-17   0.21
 0.1   -0.2          -0.1

julia> P' * pinv(QRX.R) * QRX.Q'
2×3 MMatrix{2, 3, Float64, 6} with indices SOneTo(2)×SOneTo(3):
 0.21   3.29719e-17   0.21
 0.1   -0.2          -0.1
```

lukem12345:
@jpfairbanks This PR is ready to review.

src/DiscreteExteriorCalculus.jl
```diff
@@ -801,7 +801,14 @@ function ♯_mat(s::AbstractDeltaDualComplex2D, ::LLSDDSharp)
       end
     # TODO: Move around ' as appropriate to minimize transposing.
     X = stack(de_vecs)'
-    LLS = pinv(X'*(X))*(X')
+    # See: https://arxiv.org/abs/1102.1845
+    QRX = qr(X, ColumnNorm())
+    P = zeros(MMatrix{length(QRX.p),length(QRX.p), pt.parameters[2], length(QRX.p)*2})
+    for (i,j) in zip(1:size(P,1),QRX.p)
+      P[i,j] = one(pt.parameters[2])
+    end
```
jpfairbanks, line 809: This code is just making a permutation matrix for QRX.p with the right element type, right? 

I think you can just go to the next line and index `LLS` with the permutation as a Vector{Int} one of the following 4 options should work depending on if you want to permute by rows or columns:

1. `LLS[QRX.p, :]` permutes rows by p
1. `LLS[:, QRX.p]` permutes columns by p
1. `LLS[invperm(QRX.p), :]` permutes rows by inverse of p
1. `LLS[:, invperm(QRX.p)]`

jpfairbanks (changes requested):

jpfairbanks:
Instead of creating a permutation matrix and multiplying, you can just index into the rows of the matrix.

```julia

julia> I(4)
4×4 Diagonal{Bool, Vector{Bool}}:
 1  ⋅  ⋅  ⋅
 ⋅  1  ⋅  ⋅
 ⋅  ⋅  1  ⋅
 ⋅  ⋅  ⋅  1

julia> I(4)[[2,1,3,4], :]
4×4 Matrix{Bool}:
 0  1  0  0
 1  0  0  0
 0  0  1  0
 0  0  0  1
```

lukem12345:
No problem. I've performed the operation in place now. This PR is ready for review.

lukem12345:
Perhaps it's just too late in the day, but I've gone ahead and replaced the calculation of the `pinv` with [Julia's native one](https://github.com/JuliaLang/julia/blob/baca8baea0e62f06530a9640153d793a69eba7f7/stdlib/LinearAlgebra/src/dense.jl#L1559) which uses the SVD decomposition. It feels a little wrong to be rolling our own pseudo-inverse implementation inside the CombinatorialSpaces library.

jpfairbanks:
Well, at least we aren't forming the normal equations anymore. SVD of X and using the orthogonal factors to compute the pinv is probably the most accurate thing you can do in general problems. 

We should still revisit our design choice to use a matrix representation instead of a LinearOperator callable struct.

PR #97, opened 2024-06-11
Upstream the averaging operator

lukem12345:
This is a small PR for converting 0-forms to 1-forms by taking the mean of the data stored on the faces of an edge.

`avg_01_mat` computes a matrix that performs this averaging operation.
`avg_01` constructs and multiplies the given 0-form by that matrix.
Aliases using unicode subscripts are given.

This functionality is quite old but has never been explicitly upstreamed.

### DIFF
--- a/src/DiscreteExteriorCalculus.jl
+++ b/src/DiscreteExteriorCalculus.jl
@@ -23,7 +23,8 @@ export DualSimplex, DualV, DualE, DualTri, DualChain, DualForm,
   vertex_center, edge_center, triangle_center, dual_triangle_vertices, dual_edge_vertices,
   dual_point, dual_volume, subdivide_duals!, DiagonalHodge, GeometricHodge,
   subdivide, PPSharp, AltPPSharp, DesbrunSharp, LLSDDSharp, de_sign,
-  ♭♯, ♭♯_mat, flat_sharp, flat_sharp_mat
+  ♭♯, ♭♯_mat, flat_sharp, flat_sharp_mat,
+  avg₀₁, avg_01
 
 import Base: ndims
 import Base: *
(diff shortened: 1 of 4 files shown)


src/DiscreteExteriorCalculus.jl
```diff
@@ -1474,6 +1475,42 @@ const flat_sharp = ♭♯
...
+
+Given a 0-form, this matrix computes a 1-form by taking the mean of value stored on the faces of each edge.
+
+This matrix can be used to implement a wedge product: `(avg₀₁(s)*X) .* Y` where `X` is a 0-form and `Y` a 1-form, assuming the center of an edge is halfway between its endpoints.
+
+See also [`avg₀₁`](@ref).
+"""
+function avg₀₁_mat(s::HasDeltaSet)
+  I = Vector{Int64}()
+  J = Vector{Int64}()
```
GeorgeR227, line 1488: Should have these `Int64`s be `Int32`s to better match our current code. Even better would be to add an option to choose with `Int32` being default.

src/DiscreteExteriorCalculus.jl
```diff
@@ -1474,6 +1475,42 @@ const flat_sharp = ♭♯
...
+  V = Vector{Float64}()
```
GeorgeR227, line 1489: Don't fix this to `Float64`. Let the float type depend on the mesh data.

src/DiscreteExteriorCalculus.jl
```diff
@@ -1474,6 +1475,42 @@ const flat_sharp = ♭♯
...
+  for e in 1:ne(s)
+    append!(J, [s[e,:∂v0],s[e,:∂v1]])
+    append!(I, [e,e])
+    append!(V, [0.5, 0.5])
```
GeorgeR227, line 1493: Benchmark allocations for this matrix. You'll probably have to either make these static vectors or just `push!` twice.

GeorgeR227 (changes requested):

src/DiscreteExteriorCalculus.jl
lukem12345: I've gone ahead and written a speedy version that makes use of the exterior derivative operator.

lukem12345:
Future work could take advantage of the given trick to return an averaging operator from 0-forms to 2-forms.

PR #100, opened 2024-06-15
3D Mesh IO

### LINKED ISSUE #99: Constructing large 3D meshes
PR #60 added support for 3D dual complexes, building off PR #57 which added support for 3D simplicial sets.

Currently, we can thoroughly test 2D operators on meshes constructed programmatically or read in from .obj files.

In 3D, we can construct meshes by gluing single tetrahedra, or using the [`glue_sorted_tet_cube!`](https://algebraicjulia.github.io/CombinatorialSpaces.jl/dev/simplicial_sets/#CombinatorialSpaces.SimplicialSets.glue_sorted_tet_cube!-Tuple{HasDeltaSet3D,%20Vararg{Int64,%208}}) function, which adds tesselated cubes to a mesh.

- [ ] We should find a way to load in 3D tesselated spaces from whichever file formats are convenient. Hopefully we can keep using `GeometryBasics.Mesh` objects
- [ ] We should add a helper function for creating meshes of many more tetrahedra, analagous to the [`triangulated_grid`](https://algebraicjulia.github.io/CombinatorialSpaces.jl/dev/meshes/#CombinatorialSpaces.Meshes.triangulated_grid-NTuple{5,%20Any}) function.

This will enable larger scale tests of our differential operators in 3D.

lukem12345:
Close #99 

For the time being, we can use `TetGen` on 3D Mesh objects from `GeometryBasics`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -58,6 +58,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TetGen = "c5d3f3f7-f850-59f6-8a2e-ffc6dc1317ea"
 
 [targets]
-test = ["CairoMakie", "CUDA", "Pkg", "Random", "Test", "Statistics"]
+test = ["CairoMakie", "CUDA", "Pkg", "Random", "Test", "Statistics", "TetGen"]
--- a/Project.toml
+++ b/Project.toml
@@ -48,6 +48,7 @@ MeshIO = "0.4"
 Reexport = "0.2, 1.0"
 SparseArrays = "1.9"
 StaticArrays = "0.12, 1.0"
+TetGen = "1.5"
 julia = "1.10"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,6 @@ MeshIO = "7269a6da-0436-5bbc-96c2-40638cbb6118"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
-TetGen = "c5d3f3f7-f850-59f6-8a2e-ffc6dc1317ea"
 
 [weakdeps]
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
--- a/Project.toml
+++ b/Project.toml
@@ -20,6 +20,7 @@ MeshIO = "7269a6da-0436-5bbc-96c2-40638cbb6118"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+TetGen = "c5d3f3f7-f850-59f6-8a2e-ffc6dc1317ea"
 
 [weakdeps]
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
--- a/src/MeshInterop.jl
+++ b/src/MeshInterop.jl
@@ -10,7 +10,7 @@ using FileIO, MeshIO, GeometryBasics
 
 using Catlab.CategoricalAlgebra: copy_parts!
 using ..SimplicialSets, ..DiscreteExteriorCalculus
-import ..SimplicialSets: EmbeddedDeltaSet2D
+import ..SimplicialSets: EmbeddedDeltaSet2D, EmbeddedDeltaSet3D
 
 # Export meshes
 ###############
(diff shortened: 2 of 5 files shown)


Project.toml
```diff
@@ -56,6 +58,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TetGen = "c5d3f3f7-f850-59f6-8a2e-ffc6dc1317ea"
```
GeorgeR227, line 61: How much does adding TetGen impact load/precompile times?

Project.toml
```diff
@@ -56,6 +58,7 @@ Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
...
 
 [targets]
-test = ["CairoMakie", "CUDA", "Pkg", "Random", "Test", "Statistics"]
+test = ["CairoMakie", "CUDA", "Pkg", "Random", "Test", "Statistics", "TetGen"]
```
GeorgeR227, line 64: If TetGen is part of the `deps` does it need to be part of the test dependencies?

src/MeshInterop.jl
```diff
@@ -10,57 +10,115 @@ using FileIO, MeshIO, GeometryBasics
...
+end
+
 # Import meshes
 ###############
 
-""" Construct EmbeddedDeltaSet2D from Mesh object
+# Given a vector, return an endofunction that picks out the "unique" subset,
+# and the indices of first occurences of the entries in that endofunction.
+# The second return value is equivalent to:
+# `unique(i -> ind_map[i], eachindex[ind_map])`.
```
GeorgeR227, line 84: Give a small example of what this does. I was a bit confused by what was meant by "unique" subset.

src/MeshInterop.jl
```diff
@@ -10,57 +10,115 @@ using FileIO, MeshIO, GeometryBasics
...
+function uniqueperm(coords::AbstractVector{T}, force_unique::Bool) where T
+  uniques = 1:length(coords)
+  ind_map = 1:length(coords)
+  if force_unique
+    s = Dict{T,Int}()
+    i = 1
+    uniques = Vector{Int}()
```
GeorgeR227, line 91: Is this type-stable? It seems like `uniques` can be either a `UnitRange{Int64}` or a `Vector{Int}`.

src/MeshInterop.jl
```diff
@@ -70,7 +128,34 @@ function EmbeddedDeltaSet2D(m::GeometryBasics.Mesh; force_unique=false)
...
+since some file formats reference points by location.
+"""
+function EmbeddedDeltaSet3D(m::GeometryBasics.Mesh; force_unique=false)
+  coords = metafree.(coordinates(m))
+  ind_map, uniques = uniqueperm(coords, force_unique)
+  tets = faces(m)
+  s = EmbeddedDeltaSet3D{Bool, eltype(coords)}()
+  add_vertices!(s, length(uniques), point=coords[uniques])
+  for tet in tets
+    tet = ind_map[convert.(Int64, tet)]
```
GeorgeR227, line 148: Why does this conversion need to be done?

src/Meshes.jl
```diff
@@ -356,5 +356,19 @@ function right_scalene_unit_hypot()
...
+function single_tetrahedron()
+  pnts = Point3D[
+    (0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (0.0, 0.0, 1.0)]
+  primal_s = EmbeddedDeltaSet3D{Bool, Point3D}()
+  add_vertices!(primal_s, 4, point=pnts)
+  glue_sorted_tetrahedron!(primal_s, 1:4...)
+  s = EmbeddedDeltaDualComplex3D{Bool, Float64, Point3D}(primal_s)
+  subdivide_duals!(s, Barycenter())
+  (primal_s, s)
+end
```
GeorgeR227, line 372: This should be more type generic since we'd want to use both `Float32` and `Float64`. This can be an issue since it seems that other similar functions need the same change.

src/MeshInterop.jl
lukem12345: Map, `f`, from the original vector, `V`, to `unique(V)`, such that for all indices `i` in `V`, `f(i)==v`.

Project.toml
lukem12345: Just need it for the test

Project.toml
lukem12345: I assumed that the UUID of the package has to be present in the TOML somewhere

src/MeshInterop.jl
lukem12345: It is analogous to `sortperm`.

src/MeshInterop.jl
lukem12345: Yeah I can check. But resolving it by calling collect on the range feels wasteful.

src/MeshInterop.jl
lukem12345: Was in the 2D conversion code, so I assumed was for random parsing reason.

Project.toml
GeorgeR227: True but `TetGen`'s UUID is already present in the `extras`. If you want to keep a `compat` on it then add it to weakdeps.

src/Meshes.jl
lukem12345: These functions were just intended for use in tests. I didn't find it worthwhile at the time to test Float32 for the tests that use this mesh, but I suppose that they can be made a bit more customizable.

Project.toml
lukem12345: Yeah it's not supposed to be in deps.

src/MeshInterop.jl
GeorgeR227: But in that case this isn't a permutation anymore since it's not an isomorphism. Change the name to reflect that and maybe add a tiny example for this in the docstring.

src/MeshInterop.jl
GeorgeR227: Is `faces` our own code or for `GeometryBasics`? If it's theirs then maybe namespace it as such since this conversion makes it seems we're working on different assumptions.

lukem12345: It's a GeometryBasics thing.

GeorgeR227 (approved):

PR #108, opened 2024-10-03
Clean CUDA operators

lukem12345:
In advance of adding more CUDA versions and wrappers of DEC operators (possibly based off of [KernelAbstractions.jl](https://github.com/JuliaGPU/KernelAbstractions.jl)), it is necessary to clean up the existing code.

For example, the [restrictions](https://github.com/JuliaGPU/GPUCompiler.jl/issues/597) on dispatching CUDA.jl kernel functions led to some redundant code, now deleted. The dispatch of the regular and inverse Hodge stars was also not optimal (, not exploiting the abstract `DiscreteHodge` type for example). Docstrings have also been cleaned. The docstrings for the internal functions to the wedge product are no longer exposed.

### DIFF
--- a/ext/CombinatorialSpacesCUDAExt.jl
+++ b/ext/CombinatorialSpacesCUDAExt.jl
@@ -1,6 +1,7 @@
 module CombinatorialSpacesCUDAExt
 
 using CombinatorialSpaces
+using CombinatorialSpaces.DiscreteExteriorCalculus: DiscreteHodge
 using GeometryBasics
 using CUDA
 using CUDA.CUSPARSE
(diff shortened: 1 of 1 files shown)


ext/CombinatorialSpacesCUDAExt.jl
```diff
@@ -12,176 +13,184 @@ Point3D = Point3{Float64}
...
+# Compute with a preallocated wedge product via CUDA in-place.
+function dec_c_wedge_product!(::Type{Tuple{j,k}}, res, α, β, wedge_cache, ::Type{Val{:CUDA}}) where {j,k}
+  # Manually dispatch, since CUDA.jl kernels cannot.
+  kernel = if (j,k) == (0,1)
+    dec_cu_ker_c_wedge_product_01!
+  elseif (j,k) == (0,2)
+    dec_cu_ker_c_wedge_product_02!
+  elseif (j,k) == (1,1)
+    dec_cu_ker_c_wedge_product_11!
+  else
```
GeorgeR227, line 69: Need a (1, 0) and a (2, 0) case.

ext/CombinatorialSpacesCUDAExt.jl
```diff
@@ -12,176 +13,184 @@ Point3D = Point3{Float64}
...
-                                + coeffs[Int32(5), i] * f[pv[Int32(5), i]] + coeffs[Int32(6), i] * f[pv[Int32(6), i]])
+  @inbounds while i <= Int32(length(res))
+    c1, c2, c3, c4, c5, c6 = c[Int32(1),i], c[Int32(2),i], c[Int32(3),i], c[Int32(4),i], c[Int32(5),i], c[Int32(6),i]
+    p1, p2, p3, p4, p5, p6 = p[Int32(1),i], p[Int32(2),i], p[Int32(3),i], p[Int32(4),i], p[Int32(5),i], p[Int32(6),i]
+    res[i] = α[i] * (c1*f[p1] + c2*f[p2] + c3*f[p3] + c4*f[p4] + c5*f[p5] + c6*f[p6])
     i += stride
   end
-  return nothing
+  nothing
 end
```
GeorgeR227, line 103: Has the performance of this kernel been impacted by these changes, specifically benchmark times and register counts?

ext/CombinatorialSpacesCUDAExt.jl
```diff
@@ -12,176 +13,184 @@ Point3D = Point3{Float64}
...
+    res[i] =
+      (c1 * (ae2 * be1 - ae1 * be2) +
+       c2 * (ae2 * be0 - ae0 * be2) +
+       c3 * (ae1 * be0 - ae0 * be1))
     i += stride
   end
-
-  return nothing
+  nothing
 end
```
GeorgeR227, line 123: Same performance questions as the other kernel.

GeorgeR227 (changes requested):

ext/CombinatorialSpacesCUDAExt.jl
lukem12345: Are there tests for that?

lukem12345: That's doesn't correspond to any of the `p` wedge products and is not how this function is used internally.

GeorgeR227: True but this function is exposed publicly. At the very least we can assert that the larger form appears on the right. If we wanna do more, we can check the order of the forms and adjust the inputs as needed like what happens internally in Decapodes.

I don't think there are any tests for the flipped cases.

lukem12345: If I understand correctly,  the architecture of this file is copied over from `src/FastDEC` and so the same restriction is there. I'm not going to add support for the other cases in this PR to keep the scope closed. I'll just throw an error if incorrect values are passed.

ext/CombinatorialSpacesCUDAExt.jl
lukem12345: Which benchmark script were you using?

GeorgeR227: Unfortunately there is no script for benchmarking CUDA functions, only for CPU, but the idea for benchmarking is essentially the same, just using `CUDA.@time` instead. For register counts, CUDA.jl has the function `CUDA.registers` that can be run on a kernel that will compute that.

lukem12345: Ran the the wedge product tests as CUDA synced benchmarks. Equivalent

GeorgeR227 (approved):

Issue #110, opened 2024-10-04
Buildkite Down: "No Space Left on Device"


lukem12345:
The most recent buildkite job failed due to a "no space left on device error".

```quote
🚨 Error: Error setting up Job API: starting Job API server: starting socket server: listen unix /home/rc-svc-privweb-fairbanksj/.buildkite-agent/sockets/job-api/2223947-22426.sock: bind: no space left on device
```

quffaro:
Propagating this pipeline.yml on #111 to all branches will fix this issue.

PR #111, opened 2024-10-09
Adding dynamic pipelines

quffaro:
The `pipeline.yml` now points to a template pipeline in our `.github` repo. This resolves [Issue 110](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/issues/110), as the package depot for each agent is now stored on /blue/.

[Successful build](https://buildkite.com/algebraicjulia/combinatorialspaces/builds/53)

### DIFF
--- a/.buildkite/jobscript.sh
+++ b/.buildkite/jobscript.sh
@@ -1,0 +1,9 @@
+pwd; hostname; date
+
+module load julia
+
+echo "Running Tests..."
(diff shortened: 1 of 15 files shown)


lukem12345:
The output from the tests looks like it is written to a log file. Where is this and how do I view it?

Can the script be launched on the `gpu` partition so we can run all the tests

PR #123, opened 2024-11-04
Packaging up multigrid meshes

GeorgeR227:
This is meant to wrap up the code that would generate the list of dual meshes. Also leads nicely into a MultigridData constructor. This is mainly aimed for ease-of-use in Decapodes.jl.

### DIFF
--- a/src/CombinatorialSpaces.jl
+++ b/src/CombinatorialSpaces.jl
@@ -15,9 +15,9 @@ include("FastDEC.jl")
 
 @reexport using .SimplicialSets
 @reexport using .DiscreteExteriorCalculus
-@reexport using .Multigrid
-@reexport using .FastDEC
 @reexport using .MeshInterop
 @reexport using .Meshes
+@reexport using .Multigrid
+@reexport using .FastDEC
 
 end
--- a/src/CombinatorialSpaces.jl
+++ b/src/CombinatorialSpaces.jl
@@ -8,16 +8,16 @@ include("ExteriorCalculus.jl")
 include("SimplicialSets.jl")
 include("DiscreteExteriorCalculus.jl")
 include("MeshInterop.jl")
-include("FastDEC.jl")
 include("Meshes.jl")
 include("restrictions.jl")
 include("Multigrid.jl")
+include("FastDEC.jl")
 
 @reexport using .SimplicialSets
 @reexport using .DiscreteExteriorCalculus
+@reexport using .Multigrid
 @reexport using .FastDEC
 @reexport using .MeshInterop
 @reexport using .Meshes
-@reexport using .Multigrid
 
 end
--- a/src/FastDEC.jl
+++ b/src/FastDEC.jl
@@ -12,7 +12,7 @@ module FastDEC
 
 using ..SimplicialSets, ..DiscreteExteriorCalculus
 using ..DiscreteExteriorCalculus: crossdot
-using ..Multigrid: PrimitiveGeometricMapSeries, MultigridData, full_multigrid
+using ..Multigrid: AbstractGeometricMapSeries, MultigridData, full_multigrid
 
 using ACSets
 using Base.Iterators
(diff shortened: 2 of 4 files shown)


src/CombinatorialSpaces.jl
lukem12345, line 1: Does the order of these includes matter now?

src/FastDEC.jl
```diff
@@ -19,17 +19,20 @@ using KernelAbstractions
 using LinearAlgebra: cross, dot, Diagonal, factorize, norm
 using SparseArrays: sparse, spzeros, SparseMatrixCSC
 using StaticArrays: SVector, MVector
+using Krylov: cg
 
 import ..DiscreteExteriorCalculus: ∧
 import ..SimplicialSets: numeric_sign
+import ..Multigrid: PrimitiveGeometricMapSeries, MultigridData
```
lukem12345, line 26: Why `import` and not `using`?

src/FastDEC.jl
```diff
@@ -632,6 +635,11 @@ function Δᵈ(::Type{Val{1}}, s::SimplicialSets.HasDeltaSet)
   end
 end
```
lukem12345, line 637: Docstring missing

src/Multigrid.jl
```diff
@@ -68,6 +69,30 @@ function repeated_subdivisions(k,ss,subdivider)
   end
 end
 
+"""
+This struct is meant to organize the mesh data required for multigrid methods.
```
lukem12345, line 73: "meant to organize" -> "organizes"

src/Multigrid.jl
```diff
@@ -68,6 +69,30 @@ function repeated_subdivisions(k,ss,subdivider)
...
+"""
+struct PrimitiveGeometricMapSeries{D<:HasDeltaSet, M<:AbstractMatrix}
+  meshes::AbstractVector{D}
+  matrices::AbstractVector{M}
+
+  function PrimitiveGeometricMapSeries{D, M}(meshes, matrices) where {D, M}
```
lukem12345, line 79: Why is this inner constructor necessary?

src/Multigrid.jl
```diff
@@ -68,6 +69,30 @@ function repeated_subdivisions(k,ss,subdivider)
...
+    new(meshes, matrices)
+  end
+
+  function PrimitiveGeometricMapSeries(s, subdivider, levels, alg = Circumcenter())
```
lukem12345, line 83: It is unclear what the type of `s` is.

src/Multigrid.jl
```diff
@@ -68,6 +69,30 @@ function repeated_subdivisions(k,ss,subdivider)
...
+    subdivs = reverse(repeated_subdivisions(levels, s, binary_subdivision_map));
+    meshes = map(subdivs) do subdiv dom(subdiv) end
```
lukem12345, line 85: I think this `map do end` one-liner syntax is copied over from somewhere?
`meshes = map(subdiv -> dom(subdiv), subdivs)` is better
`meshes = dom.(subdivs)` is superior if it works.

src/Multigrid.jl
```diff
@@ -68,6 +69,30 @@ function repeated_subdivisions(k,ss,subdivider)
...
+    dual_meshes = map(meshes) do s dualize(s, alg) end
+    matrices = as_matrix.(subdivs)
+    PrimitiveGeometricMapSeries{typeof(first(dual_meshes)), typeof(first(matrices))}(dual_meshes, matrices)
+  end
+end
+
+finest_mesh(series::PrimitiveGeometricMapSeries) = first(series.meshes)
+
 """
 A cute little package contain your multigrid data. If there are
```
lukem12345, line 97: I think this is a tongue-in-cheek docstring from Kevin that made it through the code review. Go ahead and change the first sentence to something else.

Compare this against the docstring for `PrimitiveGeometricMapSeries`, which sounds like they do the same thing. The docstrings of these two similar sounding structures does not explain at first glance what they do differently.

Workshop the name `PrimitiveGeometricMapSeries`. I don't know what the technical connotations of "Series" are. Is there a fundamental idiom that is being used here? It's a vector that you want a special constructor for?

src/Multigrid.jl
```diff
@@ -87,6 +112,14 @@ on each grid.
 """
 MultigridData(g,r,p,s::Int) = MultigridData{typeof(g),typeof(r)}(g,r,p,fill(s,length(g)))
 
+function MultigridData(series::PrimitiveGeometricMapSeries, op::Function, s)
```
lukem12345, line 115: I am getting deja vu on this function, especially the comment "4 is the biggest row sum that occurs for triforce, this is not clearly the correct scaling".

Is this code copied-and-pasted from somewhere else in the codebase?

src/Multigrid.jl
GeorgeR227: Yeah, this code was originally at the top of the multigrid tests. My goal was to essentially put all that code into a single function. Your suggestion is good though.

src/Multigrid.jl
GeorgeR227: Yeah this is also shuffled over from the test. I'm sure there's a more principled way to handle this scaling but as of now I don't know how this factor will be decided, what it's dependent on or if it'll even be constant.

src/CombinatorialSpaces.jl
GeorgeR227: They do. FastDEC depends on Multigrid now so FastDEC needs to appear after it.

src/Multigrid.jl
GeorgeR227: Since inner constructors will remove the default constructor. But it makes me wonder if we actually need these to be inner constructors and I don't think they do.

lukem12345: Something about this feels like an antipattern for julia code

src/Multigrid.jl
GeorgeR227: I've tagged it as `HasDeltaSet`.

PR #127, opened 2024-11-20
Clean deps for new release

GeorgeR227:
These packages had no compat entries and upon further inspection were not being used. I'm removing them now since this is [blocking the triggering of a new release](https://github.com/JuliaRegistries/General/pull/119840). 

Perhaps we should also add Aqua.jl to this package to avoid this kind of issue in the further.


lukem12345:
Yeah you have to close CompatHelper PRs before triggering a release

GeorgeR227:
Ok, then I can bump MeshIO and GeometryBasics here since it seems like they're stepping on each other in the compat PRs. I'll close the PRs for AlgebraicInterfaces, LinearSolve, and RowEchelon since those have all been removed now.

GeorgeR227:
Also seems like a [test failed on Ubuntu](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/actions/runs/11939152938/job/33279050180) due to a too strict bound in `test/Multigrid.jl`, I'll loosen that up here.

lukem12345 (approved):

PR #129, opened 2024-12-03
Revert breaking changes in 0.6.8

GeorgeR227:
This PR is meant to remove breaking changes so that `0.6.9` can release without breaking changes. Then this PR can be reverted upon which `0.7.0` can be released. The reason for this is that then any compats accepting any `0.6.x` will update to this non-breaking version.

This version of CombinatorialSpaces works with Decapodes version `0.5.6`. The next version of Decapodes, likely `0.5.7` containing https://github.com/AlgebraicJulia/Decapodes.jl/pull/276, should then bump compat to only work with CombinatorialSpaces `0.7.x`.

Most changes reverted here are those made in PR https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/pull/109.


lukem12345:
@GeorgeR227 What is the status of this PR

GeorgeR227:
This can be closed now but next version of CombinatorialSpaces should be bumped to 0.7.x since the changes in the wedge functions breaks compatibility.

PR #131, opened 2025-01-17
Faster Triangulated Grid Generation

GeorgeR227:
This PR speeds up `triangulated_grid` generation by around x20 what it was previously and also adds support for `Point3{Float32}`. The order of the points is preserved from the original but the order of the triangles has changed so that they are added left to right and bottom to top.

The main gain in the speed here was using the order of the triangles to speed up the orientation generation significantly. The bottleneck now is the `glue_triangle` section which if improved could lead to another magnitude of speedups.

Below are rough benchmarks run on my local machine.

```julia
This is a 8 by 8 grid
  Old: 0.002595 seconds (9.40 k allocations: 1.011 MiB)
  New: 0.000109 seconds (1.18 k allocations: 130.406 KiB)
This is a 16 by 16 grid
  Old: 0.004915 seconds (38.88 k allocations: 3.983 MiB)
  New: 0.000340 seconds (4.20 k allocations: 453.703 KiB)
This is a 32 by 32 grid
  Old: 0.019402 seconds (215.87 k allocations: 16.721 MiB)
  New: 0.001193 seconds (15.94 k allocations: 1.592 MiB)
This is a 64 by 64 grid
  Old: 0.242466 seconds (929.23 k allocations: 68.011 MiB, 69.82% gc time)
  New: 0.003419 seconds (62.43 k allocations: 6.659 MiB)
This is a 128 by 128 grid
  Old: 0.695434 seconds (3.78 M allocations: 271.332 MiB, 60.86% gc time)
  New: 0.024708 seconds (247.52 k allocations: 24.528 MiB, 45.29% gc time)
This is a 256 by 256 grid
  Old: 1.458329 seconds (15.19 M allocations: 1.059 GiB, 26.26% gc time)
  New: 0.089756 seconds (986.28 k allocations: 98.278 MiB, 39.19% gc time)
This is a 512 by 512 grid
  Old: 5.390635 seconds (60.81 M allocations: 4.240 GiB, 19.05% gc time)
  New: 0.308000 seconds (3.94 M allocations: 396.969 MiB, 26.11% gc time)
```

### DIFF
--- a/src/Meshes.jl
+++ b/src/Meshes.jl
@@ -9,6 +9,7 @@ using GeometryBasics: Point2, Point3
 using LinearAlgebra: diagm
 
 export loadmesh, Icosphere, Rectangle_30x10, Torus_30x10, Point_Map, triangulated_grid, makeSphere
+export new_triangulated_grid
 
 Point2D = Point2{Float64}
 Point3D = Point3{Float64}
@@ -101,7 +102,7 @@ function triangulated_grid(max_x, max_y, dx, dy, point_type, compress=true)
       end
     end
   end
- 
+
   add_vertices!(s, length(coords), point = vec(coords))
 
   nx = length(0:dx:max_x)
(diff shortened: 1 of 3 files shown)


lukem12345:
Changing the default edge orientation will change downstream code that relies on flattening vector fields, so after merging we will have to update examples in Decapodes, and check on the Euler page here.

lukem12345 (approved):
We can add mesh generation to the Benchmark script at some later date.

Issue #82, opened 2024-05-15
WebGL backend for Makie not rendering plots


lukem12345:
It looks like the docs for meshes are missing the plots on this page https://algebraicjulia.github.io/CombinatorialSpaces.jl/dev/meshes/

_Originally posted by @jpfairbanks in https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/issues/80#issuecomment-2110523156_

PR #134, opened 2025-01-26
Update compats for new Makie

lukem12345:
Close #130


lukem12345:
@quffaro If this build passes, you should check that the [Documenter PR on Decapodes](https://github.com/AlgebraicJulia/Decapodes.jl/pull/296) works on this branch. This current PR allows for the most recent version of Makie, which has silently broken docs plots in the past: https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/issues/82

quffaro (approved):

PR #135, opened 2025-01-28
Primal to dual triangle interpolation

GeorgeR227:
This PR introduces `p2_d2_interpolation` which is meant to faithfully map primal Form-2 data onto dual Form-2 data. The process looks at the dual to a primal point which is composed of dual triangles from primal triangles upon which this primal point is incident. The values from these primal triangles are transferred to their dual triangle components by area weighting these dual triangle components. Components from individual primal triangles are then all summed to generate the value for the dual Form-2.

Using the inverse hodge 0 after interpolating can spread out data from primal triangles onto primal points. Similarly, using inverse hodge 2 can first map data from dual points onto primal triangles. Moving through the entire pipeline, this serves as a simple dual point to primal point interpolation.


test/DiscreteExteriorCalculus.jl
```diff
@@ -704,6 +704,79 @@ for (primal_s,s) in flat_meshes
     atol=1e-12))
 end
 
+for (s, sd) in flat_meshes
```
lukem12345, line 707: The interpolation function should get tested on an icosphere mesh too

test/DiscreteExteriorCalculus.jl
```diff
@@ -704,6 +704,79 @@ for (primal_s,s) in flat_meshes
...
+interior_points = intersect(y_interior_points, x_interior_points)
+
+primal_points = d_rect[:point]
+dual_points = d_rect[triangle_center(d_rect), :dual_point]
+
+interp_mat = d1_p1_interpolation(d_rect)
+
+sparsity = dual_derivative(1, d_rect) * dual_derivative(0,d_rect)
+
+@test interp_mat.colptr == sparsity.colptr
```
lukem12345, line 741: I figured out that the canonical way to get the sparsity pattern of a sparse matrix is:
```julia
I,J,V = findnz(interp_mat)
@test sparsity == spzeros(I,J)
```

src/FastDEC.jl
```diff
@@ -682,4 +683,13 @@ const avg_01 = avg₀₁
 """
 const avg_01_mat = avg₀₁_mat
 
+"""   d1_p1_interpolation(sd::HasDeltaSet2D; hodge=GeometricHodge())
+
+Generates a sparse matrix that converts data on dual 0-forms into data on primal 0-forms.
+This uses the [`p2_d2_interpolation`](@ref) function as an intermediate step.
+"""
+function d1_p1_interpolation(sd::HasDeltaSet2D; hodge=GeometricHodge())
```
lukem12345, line 691: I think this function is supposed to be called `d0_p0_interpolation`

GeorgeR227:
Investigating the error caused by this interpolation I've drawn some conclusions.

- For constant functions over primal triangles, the results seems to be exact, even on the boundaries.
- For linear functions, error shows up only on the boundaries in which the gradients of the function span but the results are exact on the interior. This means that for simulations that have set boundary conditions, this interpolation will work well. The positivity/negativity of the error is also dependent on if the function is increasing/decreasing in that direction. It may be possible to apply a correction factor to remedy this but I haven't yet investigated this.

The analytic function on primal points. Function is 2x + 3y.
image

The absolute error between the analytic and the interpolated analytic function from dual points.
image

- For non-linear functions, quadratic in this example, error on the boundary follows the same rule as before in that it follows the gradients. However, in this case, there is some appreciable error on the interior. Of course, if the function is dominated by either a linear or a constant term then this error will tend towards zero.

The analytic function on primal points. Function is 1/10*x^2 + 5 where x is centered in the center of the mesh.
image

The relative error between the analytic and the interpolated analytic function from dual points.
image

GeorgeR227:
I've taken the liberty of plotting the evolution of relative interior error over time for the quadratic function mentioned above. Hopefully this serves as a guide to determine how fine a mesh is needed when only tolerating a certain amount of error. Mesh is a 32x16 sized triangulated grid with grid spacing going from 4 till 0.0625 by halving.

image

PR #140, opened 2025-02-18
Faster Mesh Binary Subdivision

GeorgeR227:
This PR builds upon PR #137 by also targeting `binary_subdivision` for optimization. By making full use of the simplicial invariants in CombinatorialSpaces, the function is able to directly create the required vertices, edges and triangles instead of relying on the slower `glue_sorted_triangle!` function.

Speedups upon PR creation are already at x2 and I expect better manipulation of ACSet setting can drive further speedups.

### DIFF
--- a/src/Multigrid.jl
+++ b/src/Multigrid.jl
@@ -4,7 +4,7 @@ using GeometryBasics:Point3, Point2
 using Krylov, Catlab, SparseArrays, StaticArrays
 using ..SimplicialSets
 import Catlab: dom,codom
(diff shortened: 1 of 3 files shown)


GeorgeR227:
Now that the new, optimized function works I've been trying to tweak the code to get the most performance out of the ACSets that I can but whatever I do doesn't really seem to impact the code runtime. I think a lot of the slowdown here is still from ACSets making a lot of allocations.

I'll try a working at it a bit more but if I don't get anything substantial, I'll clean up the current code and leave it at that.

lukem12345:
It looks like you can do batch assignments instead of looping to reduce the total number of times set subpart is called.

GeorgeR227:
I've tried that myself and didn't notice any kind of speed up but perhaps I was just doing the batch update wrong.

PR #144, opened 2025-02-24
Reorganize multigrid API

lukem12345:
This first commit rearranges the source code. The multiple dispatch/ "Julian nonsense" code was scattered in with the definitions of subdivision schemes and multigrid algorithms. The readability is improved.

### DIFF
--- a/src/Multigrid.jl
+++ b/src/Multigrid.jl
@@ -1,12 +1,22 @@
 module Multigrid
+
 using CombinatorialSpaces
-using GeometryBasics:Point3, Point2
+using GeometryBasics: Point3, Point2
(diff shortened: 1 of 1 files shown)


lukem12345:
One thing that could be done is to "lean in" to the dom/ codom idiom by using these as pointers in a doubly-linked list. This list can be traversed during the multigrid algorithm.

lukem12345:
Since this PR just moves around the code internally, (and resets the `is_simplicial_complex` check to as it was at the last release), I will merge this PR as is, and trigger a patch release. The recent PRs which introduce an AbstractSubdivisionScheme type do not break compatibility, as they translate to prior calls.

PR #153, opened 2025-04-25
Remove DataMigrations dependency

lukem12345:


GeorgeR227:
We should also remove any commented code that was meant to run with DataMigrations, but otherwise LGTM.

GeorgeR227 (approved):

PR #154, opened 2025-04-29
Mesh Optimization

lukem12345:
This PR implements a mesh optimization algorithm. The loss function to be minimized is a measure of the "equilaterality" of a triangle.

Initial dome mesh:
IMG_1569

Optimized dome mesh:
IMG_1570

Initial dome mesh with even error distribution:
IMG_1571

Optimized dome mesh with error concentrated to a portion of the domain:
IMG_1572

Error of conservation of mass. Orange is the original mesh, and blue is the optimized mesh:
IMG_1573


lukem12345:
This PR introduces a mesh optimization interface that should be generic enough for plausible future optimization algorithms. An `AbstractionMeshOptimizer` type is provided, which `SimulatedAnnealing` subtypes. This struct contains the parameters to be used by the `optimize_mesh!` function. Wrapping and unwrapping these (keyword) arguments is handled with the `Parameters` dependency.

lukem12345:
A follow up PR should make introduce this as a docs page.

test/MeshOptimization.jl
```diff
@@ -0,0 +1,198 @@
...
+eqs = optimize_mesh!(s,
+  SimulatedAnnealing(ϵ=1e-3, epochs=25, anneal=false, jitter3D=true, spherical=true))
+
+@test eqs[end] < eqs[begin]
+@test all(<(1e-3), abs.(diff(eqs)))
+# The rate of improvement tends to decrease:
+@test 0 < mean(diff(diff(eqs)))
+end
+
+let # Test the effect of mesh optimization on simple heat equation dynamics.
```
GeorgeR227, line 73: These comments should be removed if they're not planned to be part of the tests.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+```
+
+where ``E_0, E_1, E_2 = |e_0, e_1, e_2|_1``, and ``\bar{E}`` is their average.
+
+0 implies all edge lengths are equal and the triangle is equilateral.
+"""
+function equilaterality(e0, e1, e2)
+  e0, e1, e2 = normalize([e0, e1, e2], 1)
+  avg = (e0+e1+e2)/3
+  (e0 - avg)^2 + (e1 - avg)^2 + (e2 - avg)^2
```
GeorgeR227, line 40: This should use the generic `norm` function with the type of norm (L1, L2, etc.) being a kwarg.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+using Parameters
+using Random
+using SparseArrays
+using StaticArrays: SVector
+
+export optimize_mesh!, AbstractMeshOptimizer, SimulatedAnnealing, equilaterality
+
+# Given a 2D simplicial set and its 2nd boundary matrix,
+# return the edge lengths per triangle.
+function edge_lengths(s, ∂₂)
```
GeorgeR227, line 16: Could this function instead pull the edge lengths directly from the ACSet instead of using the boundary matrix? It doesn't seem like the boundary orientations matter here.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+# Compute the equilaterality of each triple of edge lengths.
+equilateralities(els) = map(x -> equilaterality(x...), els)
+
+"""    function equilaterality(s::HasDeltaSet2D, ∂₂=∂(2,s))
+
+Compute the sum of squared equilaterality of each triangle over a mesh.
+
+0 implies all triangles are equilateral.
+"""
+function equilaterality(s::HasDeltaSet2D, ∂₂=∂(2,s))
```
GeorgeR227, line 52: This might be recomputing the boundary matrix every time the function is called.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+
+`epochs`: The number of times to iterate over each point in the mesh. (Defaults to `100`)
+
+`hold_boundaries`: Whether to hold the boundaries of the mesh fixed. (Defaults to `true`)
+
+`anneal`: Whether to accept jitters than increase the cost function. (Defaults to `true`)
+
+`jitter3D`: Whether to jitter in the z-dimension, too. (Defaults to `false`)
+
+`spherical`:  Whether to constrain the new jittered point to lie on the unit sphere. (Defaults to `false`)
```
GeorgeR227, line 74: This should be paired with a `radius` kwarg so it isn't just limited to the unit sphere. `1` could be the default radius.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+function optimize_mesh!(s::HasDeltaSet2D, alg::SimulatedAnnealing)
+  @unpack_SimulatedAnnealing alg
+  ∂₂ = ∂(2,s)
+  int = interior(Val{0}, s)
+  cooling_schedule = range(0.05, .001; length=epochs)
+  map(1:epochs, cooling_schedule) do epoch, temperature
+    # TODO: You could vectorize (with a MVN) instead of iterating over points.
+    for v in (hold_boundaries ? int : vertices(s))
+      jitter = jitter3D ?
+        Point3d(randn(3) * ϵ...) :
```
GeorgeR227, line 113: This might not work if the mesh uses `Point2d` and not sure what happens if the mesh uses `Float32`.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+      temp_eq = cost(s, ∂₂)
+      # Accept this change, or undo it.
+      jump_anyway = anneal && (rand() < temperature)
+      if temp_eq < orig_eq || jump_anyway
+        s[v, :point] = jittered
+      else
+        s[v, :point] = original
+      end
+    end
+    epoch % 100 == 0 && @debug "Cost at epoch $(epoch): $(cost(s, ∂₂))"
```
GeorgeR227, line 129: Number of epochs before print should be a kwarg like `verbosity`.

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+  hold_boundaries::Bool = true
+  anneal::Bool          = true
+  jitter3D::Bool        = false
+  spherical::Bool       = false
+  cost::Function        = equilaterality
+end
+
+optimize_mesh!(s::HasDeltaSet2D) = optimize_mesh!(s, SimulatedAnnealing())
+
+# TODO: Optim.jl exports optimize!. Does that matter?
```
GeorgeR227, line 92: What is the status of this and the other TODOs?

src/MeshOptimization.jl
```diff
@@ -0,0 +1,135 @@
...
+
+Note that the selection probability is directly calculated (without exp), and does not depend on the magnitude of error improvement.
+
+See also: [`SimulatedAnnealing`](@ref).
+"""
+function optimize_mesh!(s::HasDeltaSet2D, alg::SimulatedAnnealing)
+  @unpack_SimulatedAnnealing alg
+  ∂₂ = ∂(2,s)
+  int = interior(Val{0}, s)
+  cooling_schedule = range(0.05, .001; length=epochs)
```
GeorgeR227, line 108: This cooling schedule should be a function provided by the user that accepts the current epoch and returns a temperature.

GeorgeR227 (changes requested):

src/MeshOptimization.jl
lukem12345: Which?

src/MeshOptimization.jl
lukem12345: I agree. But we should note that the more things are put into the parameters, the more breaking changes when we want to change the internals of this function.

src/MeshOptimization.jl
lukem12345: Checked at REPL ✅

src/MeshOptimization.jl
lukem12345: I've abstracted away the point type, but I will hold off on Float32 due to issues with epsilons being to small in practice etc.

src/MeshOptimization.jl
lukem12345: I see what you mean, but since the function is now parameterized by the algorithm struct, I will hold off on doing this in src.

PR #155, opened 2025-05-12
Used GeometryBasics Point2/3d

GeorgeR227:
This PR changes the internals to use GeometryBasics' own `Point2/3d` instead of our `Point2/3D`. No change in behavior is expected since they have the same underlying type, namely `Point2/3{Float64}`. 

In addition, `Point2/3` and `Point2/3d` will also be exported directly now instead of relying on the user to define them. For the time being, `Point2/3D` will also be exported to avoid breaking changes, although they should eventually be dropped in favor of `Point2/3d`.


lukem12345 (approved):
Waiting for the next CATlab release and checking for `Point{3` syntax should be good!

GeorgeR227:
No, the change to imports are just to temporarily deal with the GATlab/Catlab issues. Once those are resolved, the imports should be able to be reverted.

PR #162, opened 2025-05-23
Updated multigrid docs

jpfairbanks:
This PR adds new content about subdividing meshes, and the Multigrid data structures. It includes autodocs for the multigrid module.


lukem12345:
@copilot rebase if necessary, otherwise no-op.

Issue #113, opened 2024-10-12
Continue kernel-izing operators


lukem12345:
Prior to PR #109, CombinatorialSpaces contained two copies - one for CPU, another for CUDA - of a kernel for each of three wedge products. In PR #109, these 6 kernel instances were reduced to 2 by merging them behind a single kernel abstraction: namely `@kernel` from KernelAbstractions.jl.

For example, this:
```julia
function dec_c_wedge_product!(::Type{Tuple{1,1}}, wedge_terms, α, β, val_pack)
    e, coeffs, simples = val_pack
    @inbounds for i in simples
        ae0, ae1, ae2 = α[e[1, i]], α[e[2, i]], α[e[3, i]]
        be0, be1, be2 = β[e[1, i]], β[e[2, i]], β[e[3, i]]
        c1, c2, c3 = coeffs[1, i], coeffs[2, i], coeffs[3, i]
        wedge_terms[i] = (c1 * (ae2 * be1 - ae1 * be2) + c2 * (ae2 * be0 - ae0 * be2) + c3 * (ae1 * be0 - ae0 * be1))
    end
    return wedge_terms
end
```

and this:
```julia
function dec_cu_ker_c_wedge_product_11!(res, α, β, wedge_cache)
  e, c = wedge_cache[1], wedge_cache[2]
  i = (blockIdx().x - Int32(1)) * blockDim().x + threadIdx().x   
  stride = gridDim().x * blockDim().x
  @inbounds while i <= Int32(length(res))
    e0, e1, e2 = e[Int32(1), i], e[Int32(2), i], e[Int32(3), i]
    c1, c2, c3 = c[Int32(1), i], c[Int32(2), i], c[Int32(3), i]
    ae0, ae1, ae2 = α[e0], α[e1], α[e2]
    be0, be1, be2 = β[e0], β[e1], β[e2]
    res[i] = (c1 * (ae2 * be1 - ae1 * be2) + c2 * (ae2 * be0 - ae0 * be2) + c3 * (ae1 * be0 - ae0 * be1))
    i += stride
  end
  nothing
end
```

became that:

```julia
@kernel function wedge_kernel_11!(res, @Const(α), @Const(β), @Const(e), @Const(c))
  i = @index(Global)
  e0, e1, e2 = e[Int32(1), i], e[Int32(2), i], e[Int32(3), i]
[...]

PR #163, opened 2025-05-25
Revisit reference wedge products

lukem12345:
Currently, we have two implementations of the 1-1 wedge product: A reference implementation, and a performance-oriented implementation.

However, the reference implementation was written with some compromises in pedagogy in favor of optimizing data extraction from the ACSet.

So, I have streamlined the definition using extant higher-level helper functions. This ends up adhering more closely to the mathematical definition of this operator.

### DIFF
--- a/test/DiscreteExteriorCalculus.jl
+++ b/test/DiscreteExteriorCalculus.jl
@@ -13,7 +13,7 @@ using GeometryBasics: Point, QuadFace, MetaMesh
 using LinearAlgebra: Diagonal, mul!, norm, dot, cross
 using SparseArrays
 using StaticArrays
-using Statistics: mean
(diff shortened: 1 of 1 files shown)


lukem12345:
I used the primal cell - dual cell intersection computation in eef1f0c as an idiom for a new (reference) implementation of the 2-1 wedge product.

This should ultimately be used to guide the implementation of a `FastDEC` version of this operator as established in PR #109 and described in issue #113.

lukem12345:
The intersection computation performed in eef1f0c works when the output of the wedge product is a top-level ($n$-volume) form. Since this fails for e.g. a 1-1 wedge product in 3D, I've added back the prior computation in 5b23ab8 .

PR #169, opened 2025-06-09
3D Parallelepiped Mesh

GeorgeR227:
This PR is meant to add a 3D parallelepiped mesh that uses TetGen under the hood for easy subdivision. An example of such a mesh is shown below. Note the different side lengths set as well as the shift in the top face relative to the bottom face.
tetmesh


GeorgeR227:
I can see the point that TetGen.jl is really only used for 3D meshing but in my experience it's a very small download. Most of its dependencies we already use and the only artifact it needs is TetGen. So including it in our main dependencies shouldn't incur a noticeable slowdown in precompilation.

Plus, if we made it into a package extension then this `parallelepiped` function, along with any others in the future depending on TetGen, would need to have a stub defined when TetGen.jl isn't loaded. Not a major problem but it's not nice software in my eyes.

lukem12345 (approved):
LGTM. I'll move the TetGen README mesh into src in a followup PR.

PR #172, opened 2025-06-12
3D FastDEC Operators

GeorgeR227:
This PR is meant to introduce general support for 3D meshes among various optimized DEC operators, namely those in FastDEC.jl, as well as various other helper functions, such as `eval_constant_primal_form` and others. 

As a minimum, the following operators should be implemented:

- [x] Exterior derivative for 2-forms to 3-forms
- [x] Dual exterior derivative for dual 2-forms to dual 3-forms
- [x] All Hodge stars for 3D space (DiagonalHodge only)
- [x] Wedge product for primal  2-1 forms

### DIFF
--- a/src/FastDEC.jl
+++ b/src/FastDEC.jl
@@ -732,8 +732,14 @@ function dec_inv_hodge_star(::Type{Val{1}}, sd::EmbeddedDeltaDualComplex2D, ::Ge
   x -> hdg_lu \ x
 end
 
+dec_inv_hodge_star(::Type{Val{0}}, sd::EmbeddedDeltaDualComplex3D, ::GeometricHodge) =
+  dec_inv_hodge_star(Val{0}, sd, DiagonalHodge())
+
+dec_inv_hodge_star(::Type{Val{3}}, sd::EmbeddedDeltaDualComplex3D, ::GeometricHodge) =
+  dec_inv_hodge_star(Val{3}, sd, DiagonalHodge())
+
 dec_inv_hodge_star(::Type{Val{j}}, sd::EmbeddedDeltaDualComplex3D, ::GeometricHodge) where {j} =
-  @error "The Geometric Hodge star in 3D has not yet been implemented. Please use the Diagonal Hodge star instead."
+  @error "The Geometric Hodge star in 3D for 1-forms and 2-forms has not yet been implemented. Please use the Diagonal Hodge star instead."
 
 # Interior Product and Lie Derivative
 #------------------------------------
@@ -886,7 +892,16 @@ Generates a sparse matrix that converts data on dual 0-forms into data on primal
 This uses the [`p2_d2_interpolation`](@ref) function as an intermediate step.
 """
 function d0_p0_interpolation(sd::HasDeltaSet2D; hodge=GeometricHodge())
-  return dec_inv_hodge_star(0, sd, hodge) * p2_d2_interpolation(sd) * dec_inv_hodge_star(2, sd, hodge)
+  return SparseMatrixCSC(dec_inv_hodge_star(0, sd, hodge)) * p2_d2_interpolation(sd) * SparseMatrixCSC(dec_inv_hodge_star(2, sd, hodge))
+end
+
+"""   d0_p0_interpolation(sd::HasDeltaSet3D; hodge=GeometricHodge())
+
+Generates a sparse matrix that converts data on dual 0-forms into data on primal 0-forms.
+This uses the [`p3_d3_interpolation`](@ref) function as an intermediate step.
+"""
+function d0_p0_interpolation(sd::HasDeltaSet3D; hodge=GeometricHodge())
+  return SparseMatrixCSC(dec_inv_hodge_star(0, sd, hodge)) * p3_d3_interpolation(sd) * SparseMatrixCSC(dec_inv_hodge_star(3, sd, hodge))
 end
 
 end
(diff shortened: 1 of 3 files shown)


GeorgeR227:
Yes, that's right.

test/DiscreteExteriorCalculus.jl
```diff
@@ -740,9 +751,9 @@ for (primal_s,s) in flat_meshes
   u = eval_constant_primal_form(s, u_def)
   u_star = DualForm{1}(hodge_star(1,s) * u)
 
-  v_def = SVector{3,Float64}(-g,f,0)
-  v = eval_constant_primal_form(s, v_def)
-  dec_hodge_star(2,s) * ∧(s, u, v)
+  # v_def = SVector{3,Float64}(-g,f,0)
```
lukem12345, line 754: These can be deleted.

test/DiscreteExteriorCalculus.jl
```diff
@@ -752,11 +763,11 @@ end
 
 # Constant interpolation on flat meshes with boundaries
 for (s, sd) in flat_meshes
-  ex_1 = sign(2, sd)
+  # ex_1 = sign(2, sd)
```
lukem12345, line 766: Same deletion here.

test/DiscreteExteriorCalculus.jl
```diff
@@ -975,15 +984,13 @@ subdivide_duals!(s, Barycenter())
...
-s[:tet_orientation] = false
 
 as_vec(s,e) = (point(s, tgt(s,e)) - point(s, src(s,e))) * sign(1,s,e)
 function surface_integral_dXdY(s)
   map(triangles(s)) do tri
-    e1, e2, _ = triangle_edges(s,tri)
-    e1_vec, e2_vec = as_vec(s,e1), as_vec(s,e2)
-    (cross(e1_vec, e2_vec) * sign(2,s,tri))[3] / 2
+    _, e2, e3 = triangle_edges(s,tri)
+    e3_vec, e2_vec = as_vec(s,e3), as_vec(s,e2)
```
lukem12345, line 992: Particular reason for using e3 instead, or just happy accident?

test/Operators.jl
```diff
@@ -239,15 +249,17 @@ end
...
+  # TODO: This result should return near zero on the interior
+  # TODO: The issue might arise from a numerically singular Geometric Hodge
   for sd in [tg, rect]
     twoX = map(p -> 2*p[1], sd[sd[:tri_center], :dual_point])
     nil = Δᵈ(Val{0}, sd)(twoX)
     interior_tris = setdiff(triangles(sd), boundary_inds(Val{2}, sd))
-    @test abs(mean(nil[interior_tris])) < 1e-13
-    @test std(nil[interior_tris]) < 0.31
+    @test_broken abs(mean(nil[interior_tris])) < 1e-13
+    @test_broken std(nil[interior_tris]) < 1e-13
```
lukem12345, line 262: I would increase the error bounding on the broken test, but should be clear what to do in context.

lukem12345:
I agree, just having the API (which should just copy the existing one for 2D) will be good for now. Minor releases can optimize.

src/FastDEC.jl
```diff
@@ -132,10 +217,14 @@ function dec_c_wedge_product!(::Type{Tuple{j,k}}, res, α, β, p, c) where {j,k}
...
   elseif (j,k) == (0,2)
     wedge_kernel_02!
+  elseif (j,k) == (0,3)
+    wedge_kernel_03!
   elseif (j,k) == (1,1)
     wedge_kernel_11!
+  elseif (j,k) == (2,1)
+    wedge_kernel_21!
   else
-    error("Unsupported combination of degrees $j and $k. Ensure that their sum is not greater than the degree of the complex, and the degree of the first is ≤ the degree of the second.")
```
lukem12345, line 138: Hmm, when was this constraint on this particular method dropped?

src/FastDEC.jl
```diff
@@ -72,17 +86,61 @@ function wedge_kernel_coeffs(::Type{Tuple{1,1}}, sd::EmbeddedDeltaDualComplex2D{
...
+function wedge_kernel_coeffs(::Type{Tuple{2,1}}, sd::EmbeddedDeltaDualComplex3D{Bool, float_type, _p}) where {float_type, _p}
+  coeffs = Array{float_type}(undef, 12, ntetrahedra(sd))
+  ets = Array{Int32}(undef, 10, ntetrahedra(sd))
+  
+  for tet in tetrahedra(sd)
+    d_tets = subsimplices(3, sd, tet)
+    d_volume(tets) = sum(sd[tets, :dual_vol])
+
+    ws = map(tetrahedron_vertices(sd,tet)) do v
+      d_volume(d_tets ∩ elementary_duals(0,sd,v)) / sd[tet, :vol]
```
lukem12345, line 103: This line can be optimized on a follow up PR.

lukem12345 (approved):

src/FastDEC.jl
GeorgeR227: It's because I introduced the wedge 2-1 as the base implementation to match the wedge [defined here](https://github.com/AlgebraicJulia/CombinatorialSpaces.jl/blob/f74a97ac62cd5bd46830d4749e941dc3afacfdd0/src/DiscreteExteriorCalculus.jl#L1295-L1297). So then the degree of the first is no longer less than the second. I could implement the 1-2 as the base though.

lukem12345: Ah I see. Since this is rather internal that development effort is probably not worth it.

GeorgeR227: We can do it anyways when we go back and optimize.

PR #174, opened 2025-06-20
Add support for DelaunayTriangulation.jl

GeorgeR227:
This PR adds a small interop function for the [DelaunayTriangulation.jl](https://github.com/JuliaGeometry/DelaunayTriangulation.jl) package. A user would be expected to use DelaunayTriangulation to generate a specific 2D mesh that they desire and then use the interop function to port it to an `EmbeddedDeltaSet2D`. 

The advantages are the ability to quickly generate a mesh given only a set of desired points, have it be Delaunay (which is shown to work well in general for DEC), and allow for mesh refinement (both unconstrained and constrained). 

This PR only adds DelaunayTriangulation as an extension so there is no downside for users not using this feature.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,10 +28,12 @@ TetGen = "c5d3f3f7-f850-59f6-8a2e-ffc6dc1317ea"
 
 [weakdeps]
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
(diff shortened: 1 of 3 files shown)


lukem12345:
Try the `number_type` method from this package.

lukem12345 (approved):

PR #189, opened 2025-12-23
Cubical Complexes

GeorgeR227:
This PR introduces support for 2D cubical complexes. Namely, these are rectilinear grids that can be defined implicitly. These types of grids are useful for simulations where horizontal and vertical directions are distinguished in some way (e.g. gravity is only vertical).

The implicit mesh, at the time of this comment, only depends on an explicit list of vertex points. All other elements, including dual elements, are defined from these primal points. Vertices and quadrilaterals are given both x,y coordinate labels and sequential coordinate labels, which are given from left to right and top to bottom. 

Desired features would be:
- [x] Implicit treatment of point coordinates (explicit lists of points are replaced by functions providing coordinates)
- [x] KernelAbstractions support for all DEC operators (to support multi-threading/GPU)
- [x] Proper treatment of periodic boundary conditions (meaning automatically generated and easy to apply)
- [x] Support for distributed computing (eased since the mesh is a grid and all inform is implicit and can be easy generated)


docs/src/euler.md
```diff
@@ -70,9 +70,9 @@ function plot_dvf(sd, X; ls=1f0, title="Dual Vector Field")
   f
 end
 
-sharp_pp = ♯_mat(sd, AltPPSharp())
+sharp_pd = ♯_mat(sd, AltPPSharp())
```
lukem12345: What's going on here? It looks like a primal primal sharp is being assigned a primal dual name.

lukem12345:
@GeorgeR227 For better code management, we can create a new branch off of main, and `git cherry-pick` the appropriate commits for cubical mesh data structures and their operators. The particular physics benchmarks could be hosted in a separate repo as necessary. Thoughts?

GeorgeR227:
Organization-wise that's probably the way to go. Main reason this hasn't been done yet is cause this code is being iterated along with the physics benchmarks. Essentially, cubical code is written -> physics is written -> realize bug or need for new feature -> new cubical code is written -> ad infinitum.

So for development efficiency, the code has been stuck together with the physics. So I think once we're satisfied with the 2D benchmarks for now, I'd be happy moving the physical benchmarks elsewhere.

PR #196, opened 2026-02-20
Perform Galerkin optimization ; Clean type inference ; Improve orientation code

lukem12345:
Currently, we allocate the data needed for multigrid by explicitly creating a diagram in the category of geometric maps $$GM$$ and then we compute the restriction and prolongation operators from this. That is, we store all meshes in the series concurrently. Also, we compute the operators for each mesh by calling the constructor repeatedly, which in turn necessitates computing all dual meshes.

However, these methods have high memory requirements.

So, this PR introduces a constructor that creates a `MultigridData` struct directly, without allocating a `PrimalGeometricMapSeries`. (The diagram in $$GM$$ is traversed one morphism at a time.) This is a modest performance optimization for most workflows, but can be very impactful for the garbage collector when total available memory is an issue.

This PR also performs the so-called Galerkin optimization by calling the given constructor once on the finest mesh. This provides a much more robust improvement in time and memory. The resulting solvers sacrifice about 1 digit in the residuals on random solves.

This is related to [Decapodes PR 344](https://github.com/AlgebraicJulia/Decapodes.jl/pull/344).

### DIFF
--- a/ext/CombinatorialSpacesCUDAExt.jl
+++ b/ext/CombinatorialSpacesCUDAExt.jl
@@ -13,85 +13,85 @@ import CombinatorialSpaces: cache_wedge,
 # Wedge Product
 #--------------
 
-function cache_wedge(::Type{Tuple{m,n}}, sd::EmbeddedDeltaDualComplex1D{Bool, float_type, _p}, ::Type{Val{:CUDA}}, arr_cons=CuArray, cast_float=nothing) where {float_type,_p,m,n}
+function cache_wedge(::Val{m}, ::Val{n}, sd::EmbeddedDeltaDualComplex1D{Bool, float_type, _p}, ::Val{:CUDA}, arr_cons=CuArray, cast_float=nothing) where {float_type,_p,m,n}
   cache_wedge(m, n, sd, float_type, arr_cons, cast_float)
 end
-function cache_wedge(::Type{Tuple{m,n}}, sd::EmbeddedDeltaDualComplex2D{Bool, float_type, _p}, ::Type{Val{:CUDA}}, arr_cons=CuArray, cast_float=nothing) where {float_type,_p,m,n}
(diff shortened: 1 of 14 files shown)


lukem12345:
A quick spot benchmark on the no-Galerkin vs. Galerkin version:
```julia
julia> md_no_galerkin = @btime begin
         MultigridData(s, BinarySubdivision(), 6, sd -> ∇²(0, sd), 3, galerkin=false);
         end;
  10.101 s (107119835 allocations: 9.68 GiB)

julia> md_galerkin = @btime begin
         MultigridData(s, BinarySubdivision(), 6, sd -> ∇²(0, sd), 3, galerkin=true);
         end;
  7.469 s (80197584 allocations: 7.32 GiB)
```

lukem12345:
Should look into whether https://github.com/AlgebraicJulia/ACSets.jl/pull/138 types of things can unlock the last remaining allocs

lukem12345:
@copilot Namespace remaining calls to `collect` in source code as `Base.collect`.

lukem12345:
@copilot Use https://github.com/AlgebraicJulia/Catlab.jl/releases/tag/v0.14.4 and use the ACSet functionality exported there to test for potential remaining performance regressions. Use the most recent of Julia that it is compatible with that version. It may be necessary to undo PRs that added support for recent versions of Catlab and ACSets, but the functionality introduced in the interim should not be altered. Use git blames and search for relevant PRs when needed.

lukem12345:
Some profile results from using Catlab 0.14.4 vs. 0.17.4. The Galerkin constructor is 50% faster at the 8 subdivision level (9.8 seconds vs 20.2 seconds), but this may or may not hold at higher subdivisions due to the change in how allocations are made. (This is even without fixing the last type instability in `set_orientation!`, which is the last red-capped tower on the left.) A large caveat is that the 0.14.4 version has some `@inbounds` sprinkled in when replacing type unstable `set_subpart!` calls, so maybe that can be applied to the 0.17.4 version of the code as well.
image

